PR #132, opened 2015-02-04
Adding 'nolib' option for use tag

mvala:
Problem: autoconf needs to to search external libs like glib-2.0
following format

PKG_CHECK_MODULES([glib], [glib-2.0 >= 0.0.1],, ...

note no 'lib' prefix

With this commit one can do following in project.xml for glib-2.0

<use project = "glib" prefix = "glib-2.0" test = "" repository = ""
min_major= "0" min_minor = "0" min_patch = "1" nolib = "yes" />

### DIFF
--- a/zproject_autoconf.gsl
+++ b/zproject_autoconf.gsl
@@ -154,9 +154,15 @@ PREVIOUS_LIBS="${LIBS}"
 was_$(use.project)_check_lib_detected=no
 
 .  if (use.min_version <> '0.0.0')
+.    if defined(use.nolib)
+PKG_CHECK_MODULES([$(use.project)], [$(use.prefix) >= $(use.min_version)],,
+    [AC_MSG_ERROR([Cannot find required package for $(use.prefix). Note, pkg-config is required due to specified version >= $(use.min_version)])
+  ])
+.    else
 PKG_CHECK_MODULES([$(use.project)], [lib$(use.prefix) >= $(use.min_version)],,
     [AC_MSG_ERROR([Cannot find required package for lib$(use.prefix). Note, pkg-config is required due to specified version >= $(use.min_version)])
   ])
+.    endif
 .  else
 PKG_CHECK_MODULES([$(use.project)], [lib$(use.prefix)],,
     [


hintjens:
Can you try to use only abstractions that are obvious and natural, and
portable to all build systems?

What is the problem here? How does "nolib" solve that?

mvala:
when i set in autoconf
```
PKG_CHECK_MODULES([glib], [libglib-2.0 >= 0.0.1],, ...
```
library is not found by configure script, but when i remove "lib" so config looks like this
```
PKG_CHECK_MODULES([glib], [glib-2.0 >= 0.0.1],, ...
```
then lib is found.

"lib" is always appended in zproject_autoconf.gsl.

hintjens:
The package name is apparently not fixed across systems. Sometimes it's
libglib-2.0-0 and sometimes it's glib-2.0-0. Using a "nolib" property in
the project model is pointless.

The list of known projects is defined in zproject_known_projects.xml. We
can add support for glib there.

mvala:
Adding glib in to zproject_known_projects.xml would be good idea. Can you add it please?

hintjens:
I'd like to wait for input from someone who knows how to properly deal with
glib from autoconf...

mvala:
Yes, this is good idea. Thanks
I would also need libgtop, but this can be easily done with current version
```
<use project = "gtop" prefix = "gtop-2.0" test = "" repository = "" min_major= "0" min_minor = "0" min_patch = "1" /> 
```
Maybe this can be also added to zproject_known_projects.xml.

mvala:
@hintjens, Any news on that? I googled little bit and i see most of the time 
```
PKG_CHECK_MODULES([glib], [glib-2.0 >= 0.0.1]
```
so without lib in front glib-2.0. Can you give me hint?

jemc:
@mvala 
In the linker, how is glib linked?  With `-lglib`?

jemc:
@mvala 
After PR #133 is merged, you'll be able to override `use.linkname` and `use.libname` to whatever crazy names glib uses ;-)

This solution should preserve the sanity and consistency of the generator codebase more than introducing conditional logic like `nolib`.

mvala:
I like this better then my proposal. Thanks

PR #135, opened 2015-02-11
Problem: asciidoc: FAILED: manpage document title is mandatory

### LINKED ISSUE #131: Documentation files are not generated correctly for classes
when i run ./generate.sh documentation files (doc/*.txt) for classes are not generated correctly. Only name of the class is in the file.

hintjens:
The initial call of zproject_mkman generates incomplete man pages
which then fail with this error in asciidoc. The workaround is to
regenerate all man pages using zproject_mkman. This is not clean
nor obvious.

Solution: add "clean" target that removes .1 and .3 pages and
regenerates .txt and .doc files from source.

Fixes #131

### DIFF
--- a/zproject_docs.gsl
+++ b/zproject_docs.gsl
@@ -71,6 +71,16 @@ $(name).txt:
 $(name).txt:
 \tzproject_mkman $<
 .endfor
+clean:
+\trm -f *.1 *.3
+\tzproject_mkman \
+.for project.class where private?0 = 0
+$(name) \
+.endfor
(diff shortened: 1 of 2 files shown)


mvala:
So, do i hae to do
```
./autogen.sh
./configure
make clean
make
make install
```

Because when i didn't do
```
make clean
```
it didn't work.

hintjens:
Yes, that's how it works for now. It'd be better to make a dependency from
the source to the .txt file, perhaps.

PR #136, opened 2015-02-12
Add preludefile property to project, for CZMQ

twhittock:


hintjens:
Please scope all commits with a "Problem:" statement so we know what the origin of a change is... thanks.

Issue #138, opened 2015-02-12
Problem: Build on OS X fails


febeling:
On OSX 10.10.2 and with autotools installed via Homebrew, the build fails. 

```
$ ./autogen.sh                                                                                                                              [0:25:28]
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal -I config --force -I config
autoreconf: configure.ac: tracing
autoreconf: running: glibtoolize --copy --force
glibtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, `config'.
glibtoolize: copying file `config/ltmain.sh'
glibtoolize: putting macros in AC_CONFIG_MACRO_DIR, `config'.
glibtoolize: copying file `config/libtool.m4'
glibtoolize: copying file `config/ltoptions.m4'
glibtoolize: copying file `config/ltsugar.m4'
glibtoolize: copying file `config/ltversion.m4'
glibtoolize: copying file `config/lt~obsolete.m4'
autoreconf: running: /usr/local/Cellar/autoconf/2.69/bin/autoconf --include=config --force
autoreconf: running: /usr/local/Cellar/autoconf/2.69/bin/autoheader --include=config --force
autoheader: cannot rename /var/folders/qx/60b1pl155yv3r92hhtmg78l40000gn/T//arEjFCdm/ahNZl6F8/config.hin as src/platform.h.in: No such file or directory
autoreconf: /usr/local/Cellar/autoconf/2.69/bin/autoheader failed with exit status: 1
autogen.sh: error: autoreconf exited with status 0
```

The second autoreconf step fails, as the second last line reports. (Last error message is wrong, see #139)

It can't copy over the platform.h.in config header input file, because the src directory doesn't exist. 

If I run
```
mkdir src
```
it fails one step later when trying to read `src/Makemodule.am`. Again this is easiy solved by
```
touch src/Makemodule.am
```
as an empty file is sufficient here.

These are likely only bootstrapping problems. But they can easily trip up new users.

twhittock:
I had that happen to me when I didn't run generate.sh first. I'll make a pull request. One sec.

PR #139, opened 2015-02-12
Problem: autogen.sh: reported exit code wrong

febeling:
The second autoreconf step doesn't report back the correct exit code as
the test overrides it.

The output error mentioned error in #138 from the `autogen.sh` script is cause by this part:
```
autoreconf --install --force --verbose -I config
if [ $? -ne 0 ]; then
    echo "autogen.sh: error: autoreconf exited with status $?" 1>&2
    exit 1
fi
```
The `if [` part overrides the `$?` variable and so the echoed message contains always 0, which is misleading.

### DIFF
--- a/zproject_autoconf.gsl
+++ b/zproject_autoconf.gsl
@@ -43,8 +43,9 @@ if [ $? -ne 0 ]; then
 fi
 
 autoreconf --install --force --verbose -I config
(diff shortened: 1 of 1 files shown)


febeling:
This solves the secondary problem in #138

PR #154, opened 2015-02-22
Fixes for inconsistencies in C class header regeneration and README for installation and use.

dmeehan1968:
When the C language class header is initially generated, it includes method declarations for {class}_print() and {class}_test.  When the class api (api/{class}.xml} is subsequently created or updated, the method declarations for test and print are erroneously stripped.

README steps for installation of zproject were inaccurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,10 @@
 **<a href="#toc2-58">Sample API model</a>**
 
 **<a href="#toc2-99">Installation</a>**
-&emsp;<a href="#toc3-108">autotools</a>
+&emsp;<a href="#toc3-309">GSL</a>
+&emsp;<a href="#toc3-108">zproject</a>
 
-**<a href="#toc2-118">Generate build environment in your project</a>**
+**<a href="#toc2-118">Setup your project environment</a>**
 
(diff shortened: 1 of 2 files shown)


twhittock:
Ok this broke czmq class generation because these methods are now doubly-declared (sometimes with subtly different prototypes.)
@hintjens - I'm not sure what the expected behaviour here is - either:

1. zproject declares an implicit test and print method in the api if they are not already declared manually.
2. if you're creating an api file then you should explictly declare all methods - this goes against the implicit contructor and destructor already present.

sappo:
@twhittock you may open another pull request that reverts these changes as they did brake something or fix them if that's an option.

twhittock:
I'd prefer not to revert them - clearly they're fixing an issue @dmeehan1968 encountered. I'll make a PR for the correct behaviour once I know what the correct behaviour is. I'm leaning towards option 1 above, personally.

hintjens:
What's the simplest solution...?

twhittock:
I'll push option 1 above in a sec. Already wrote it...

PR #155, opened 2015-02-22
Problem: README.md has been modified but is generated by gitdown

sappo:
Solution: Merge changes from README.md into README.txt and generate with
gitdown

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,17 +17,16 @@
 
 **<a href="#toc2-288">Installation</a>**
 &emsp;<a href="#toc3-291">GSL</a>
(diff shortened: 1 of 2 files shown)


dmeehan1968:
Apologies for applying the edits in the wrong place.  Should I be modifying README.txt instead?

sappo:
@dmeehan1968 no worries. It would be most appriciated though if you modified the README.txt instead the next time and generate the README.md with gitdown (https://github.com/imatix/gitdown) next time :-) But it's not mandatory in any way. We're always happy to improvements on the documentation.

PR #158, opened 2015-02-23
Problem: Class constructor not generating custom arguments in header or ...

dmeehan1968:
If the class api defined custom constructor arguments, the first pass build of the source file was not generating the correct method declaration.

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -288,11 +288,10 @@ $(PROJECT.PREFIX)_EXPORT void
 struct _$(class.c_name)_t {
 };
 
-//  --------------------------------------------------------------------------
-//  Create a new $(class.c_name)
+.for class.constructor as method
 
-$(class.c_name)_t *
-$(class.c_name)_new (void)
+//  $(method.description:no,block)
+$(c_method_implementation (method):)
 {
     $(class.c_name)_t *self = ($(class.c_name)_t *) zmalloc (sizeof ($(class.c_name)_t));
     assert (self);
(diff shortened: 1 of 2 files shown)


jemc:
Just an observation regarding consistency of experience: it seems like the next logical step here would be for the first pass to generate empty stubs for *all* the specified methods for you, not just for the constructor.

If no one else tackles that change, I'll do it as soon as I have the time for it.

hintjens:
We do something like this in zproto_server_c.gsl and zproto_client_c.gsl;
perhaps you can take that code as example.

On Mon, Feb 23, 2015 at 10:09 PM, Joe Eli McIlvain <notifications@github.com

jemc:
@hintjens - yes, the zproto style of inserting new stubs on the `nth` pass of generation if they are not found in the source is also an important feature to add I think, but I see it as a second step after this one, which just involves expanding the approach @dmeehan1968 took in this PR to include all methods.

twhittock:
I was under the impression that the header was written twice in the same run, first with the hard coded stubs then with the API definition. The end result being, the user never sees the hard coded stubs if they have an API defined. This was what I fixed about a week ago by adding a .close statement in after the initial generation.

jemc:
@twhittock 
You're correct, the header is written twice on the first pass. @dmeehan1968 's "Problem:" statement in the commit message is misleading in the case of the header file, but in the case of the *.c source file the problem is real, and a bit of a nuisance in terms of user experience.

twhittock:
Add I see I missed the c file part, sorry :) makes sense to me now.

jemc:
Although I just noticed that having the formatting for implementation signature be the same as in the header as this commit does doesn't match the CLASS indentation style in existing projects, so I'll have to address that in my PR as well.

dmeehan1968:
Apologies, I had spotted the fault but decided against another PR just for whitespace.

PR #162, opened 2015-02-25
Add support for wide used integer types

bkmit:
Add support for wide used integer types [u]int(32|64)


jemc:
@bkmit 
I think the goals of zproject include trying to avoid letting these low-level C details leak upward into the model.  @hintjens has often been a watchdog on this point.  If you need these details explicitly in your project, try using the c_type attribute explicitly instead.

bkmit:
@jemc You are right! Metaprogramming is metaprogramming!

hintjens:
If you're going to add integer types to zproject then I suggest (a) stating
the problem case so we can discuss alternatives and (b) using the zproto
type system (also as used in CZMQ), which at least stays away from C-isms:

- integer (signed)
- number 1, 2, 4, 8 (unsigned byte, 16 bit, 32-bit, 64-bit)

-Pieter

PR #168, opened 2015-03-06
Fix include guards.

elfring:
The generation of include guards did not fit to [the expected naming convention of the C++ language standard](https://www.securecoding.cert.org/confluence/display/cplusplus/DCL32-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL32-CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29 "Do not use identifiers that are reserved for the compiler implementation.").

This implementation detail can be fixed by deletion of [a few underscores](http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier#answer-228797 "What are the rules about using an underscore in a C++ identifier?").

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -33,8 +33,8 @@
     =========================================================================
 */
 
-#ifndef __$(PROJECT.NAME)_H_INCLUDED__
-#define __$(PROJECT.NAME)_H_INCLUDED__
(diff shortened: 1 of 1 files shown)


jemc:
Thanks!

PR #220, opened 2015-09-03
Problem: zproject cannot handle optional libraries

sappo:
Solution: Don't throw an error if a optional library wasn't found and set a HAVE_$(LIB_LINKNAME) define if it was found. This permits to check for the library when platform.h is included in the source.


hintjens:
Does this solve the problem with libuuid?

sappo:
Yes, it should. I currently use it with libcurl which works fine.
Am 03.09.2015 15:10 schrieb "Pieter Hintjens" <notifications@github.com>:

sappo:
Okay, it doesn't ;( Because we're not able to generate an appropriate include section as it's needed for libuuid (see czmq_prelude.h). One solution that comes to mind is to add another flag which indicates not to automatically generate an include statement or never generate include statements for optional dependencies.

PR #235, opened 2015-09-17
Problem: Cannot specify uint8-64_t types in class api

sappo:
Soltion: Use the xml syntax from zproto to add uint8-64_t types

### DIFF
--- a/zproject_class_api.gsl
+++ b/zproject_class_api.gsl
@@ -39,16 +39,17 @@ function resolve_c_container (container)
     my.container.fresh ?= "0"
     my.container.fresh = conv.number (my.container.fresh)
     my.container.is_format ?= 0
-    
+
     my.container.type ?= "nothing"
     if defined (my.container.c_type)
         return
     endif
-    
+
     my.type = my.container.type
+    my.size = my.container.size? "1"
     my.c_type = ""
     my.stars = ""
-    
+
     if my.type = ""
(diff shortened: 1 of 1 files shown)


hintjens:
@sappo, nice stuff. I was trying to use these structures to generate code (JNI), and it's a little obscure. Could you have time to write down how the container structures work, for people writing binding generators?

sappo:
@hintjens once I figured that stuff out myself, I'll try to document it :-D
I'm currently writing a binding for Qt and I think I'm finally close to having a good overview about the structures.

hintjens:
Ah, cool, that will be awesome.

jemc:
Since I wrote the first two of the binding generators, I can probably explain how they work.  I'll try to do a writeup later tonight or at the end of the week.

twhittock:
I seem to remember using `XML.save` or similar to dump the document
structure after the initial passes were made - this was useful in
understanding what was going on.

jemc:
@twhittock - Great idea!  I could mention that in the writeup as well and show an example output.

@sappo - Also, you probably already saw the QML bindings, but in case you didn't they should be very helpful in creating Qt bindings (I expect they would be almost the same).

jemc:
See https://github.com/zeromq/zproject/pull/236

sappo:
@jemc - That's the reason I weren't completely lost ;)

PR #236, opened 2015-09-18
Problem: Bindings generator models are hard to understand.

jemc:
Solution: Improve consistency and documentation of the models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,14 +19,23 @@
 
 **<a href="#toc2-104">Configuration</a>**
 
-**<a href="#toc2-255">Sample API model</a>**
+**<a href="#toc2-256">Sample API model</a>**
+&emsp;<a href="#toc3-365">Supported API Model Attributes</a>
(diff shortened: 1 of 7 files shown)


jemc:
@hintjens - hopefully this helps to understand what's going on in the initial pass of the API model resolving.  I've added a combination of code comments and README material that should make the API models more approachable, as well as fixing some inconsistencies that were making the model harder to understand (now all semantic attributes are fully resolved in the first pass).

Please feel free to ask for clarification if anything is still unclear, and please fill in the gaps with your own PR if you figure out anything important that I missed.

@twhittock - I haven't used the `enum` handling feature you added, so I wasn't sure how to document it.  If you get the chance to write up some quick notes in the README, I think that would be helpful.

Let's make zproject more approachable so we can attract more code generator authors and users!

sappo:
@jemc this is really helpful, thanks :+1:

PR #237, opened 2015-09-18
Problem: zprojects that use a zproject (e.g. zyre) need used zprojects api files for their own bindings.

sappo:


sappo:
Next will be to automatically resolve depended library data types and load their api files. Once I got my prototype in shape later today.

sappo:
Okay, I didn't have the time to push it today so I'll have to postpone automatically resolving dependencies till Monday. But I got it working, so stay tuned ;)

PR #239, opened 2015-09-21
Problem: Bindings of a zproject that uses another zproject needs the dependent projects data types resolved.

sappo:
Solution: Parse the api files for foreign data types and try to find the
installed api files from the dependent project.

This way, e.g. bindings for zyre can load the necessary api files from czmq
for the data types used in the zyre api. The class api for dependencies is
loaded to: project.dependencies.class and contains the same attributes
as: project.class.

### DIFF
--- a/zproject_class_api.gsl
+++ b/zproject_class_api.gsl
@@ -134,22 +134,22 @@ function resolve_c_container (container)
         my.c_type = "$(my.type:c)_t"
         my.stars += "*"
     endif
-    
+
     if my.container.by_reference
         my.stars += "*"
     endif
-    
+
     # This attribute is resolved late because it may be influenced during type resolution
     my.container.constant ?= "0"
     my.container.constant ?= conv.number (my.container.constant)
     if my.container.constant
         my.c_type = "const " + my.c_type
     endif
-    
+
     if string.length (my.stars)
         my.c_type += (" " + my.stars)
     endif
-    
+
     my.container.c_type = my.c_type
 endfunction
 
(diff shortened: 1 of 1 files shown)


zproject_class_api.gsl
```diff
@@ -295,10 +295,49 @@ function resolve_c_class (class)
...
+# it exists.
+#
+function resolve_container_dependencies (class_name, container)
+    if my.container.c_type = "$(my.container.type:c)_t *" \
+     & !(my.class_name = my.container.type) \
+     & count (project.class, class.name = my.container.type) = 0 \
+     & count (project->dependencies.class, class.name = my.container.type) = 0
+        if file.exists ("/usr/local/bin/$(my.container.type:c).xml")
+            echo "Processing /usr/local/bin/$(my.container.type:c).xml..."
+            new_dependency = XML.load_file ("/usr/local/bin/$(my.container.type:c).xml")
```
jemc: Why `/usr/local/bin` for the API XML files?  Is there a better place to put these, since they are not executable scripts?  Maybe `/usr/local/share/zproject`?

Also, I'm thinking they should be installed into a subdirectory with the project name.  So, maybe `/usr/local/share/zproject/{project_name}/{class_name}.xml`?

If we plan to make this part of the typical `make install` process for projects, we have to be prepared to address concerns of package maintainers for various linux distros, so we should try to figure out and use some best practices here.

sappo:
@jemc honestly, I didn't put much thought into where to place the xml api files. `/usr/local/bin` was the easiest thus I used it. 

I'm totally fine if we place them in `/usr/local/share/zproject`. I'm against sub-directories tough as we cannot iterate them in gsl, at least to my knowledge.

sappo:
Okay, I did some investigation. It is actually really easy to iterate directories with gsl so I'm for sub-directories ;)

PR #242, opened 2015-09-22
Problem: API files are installed into the same directory

sappo:
Solution: Group API files by library and install them into sub-directories


jemc:
Thanks, @sappo!

PR #243, opened 2015-09-24
Problem: ztrie uses a char in its API to pass a delimiter which is not supported by zproject's API

sappo:
Solution: Add char as possible type

### DIFF
--- a/zproject_class_api.gsl
+++ b/zproject_class_api.gsl
@@ -82,6 +82,8 @@ function resolve_c_container (container)
         my.stars += "*"
     elsif my.type = "byte"
         my.c_type = "byte"
+    elsif my.type = "char"
+        my.c_type = "char"
     elsif my.type = "integer"
         my.c_type = "int"
     elsif my.type = "size"


twhittock:
what are the characteristics of 'char' ? signed byte? unsigned byte? should
we use names to describe the type, instead of talking about C types? That's
why byte is there already right?

sappo:
Wow, you ask a lot of questions :P 
unsigned char. Maybe character instead of char. A char contains textual data whereas a byte can contain anything.

hintjens:
Do you have a use case for a single char as compared to a short string? I
mean, what's the problem we're solving here ("lack of feature X" is not a
valid problem :)

sappo:
ztrie uses a char as an argument to define the delimiter.

sappo:
Example: `ztrie_new ('/');`

hintjens:
Makes sense to me.

sappo:
Shall I close this PR and open another with a commit that states the valid problem?

sappo:
Okay I managed to change the commit message ;)

PR #245, opened 2015-09-26
Problem: I would like to use liblognorm with zproject

taotetek:
Solution: Add liblognorm and it's json-c dependency to zproject.


jemc:
I've merged, but note that the rest of the list was alphabetized.

PR #262, opened 2015-10-25
Problem: code for test command line has issues

hintjens:
- empty function argument lists (use 'void')
- use of variables like i, p, etc. (formally disallowed)
- inline declarations in for loops (not standard C)
- action in for loops (move to body, it's cleaner)
- incorrect use of parenthesis (place opening { with statement)
- use of preincrement in for loops (why? we don't do that anywhere else)

Solution: fix these

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -156,7 +156,7 @@ typedef void (*testfn_t) (bool);
 typedef struct
 {
     const char *testname;
-    testfn_t     test;
+    testfn_t test;
 };
 
 //  -------------------------------------------------------------------------
@@ -165,11 +165,11 @@ typedef struct
 
 #define DECLARE_TEST(TEST) {#TEST, TEST}
 
-test_item_t all_tests[] = {
+test_item_t all_tests [] = {
 .for class
     DECLARE_TEST($(class.c_name)_test),
 .endfor
-    {0, 0} //Null terminator
+    {0, 0} // Null terminator
 };
 
 //  -------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


zproject_class.gsl
```diff
@@ -231,56 +237,50 @@ main (int argc, char *argv [])
 {
     bool verbose = false;
     test_item_t *test = 0;
-    for (int i = 1; i < argc; ++i)
-    {
-        if (streq (argv[i], "-v"))
-        {
+    for (int argn = 1; argn < argc; argn) {
```
jimklimov: Shouldn't int argn also go to a separate line away from the loop?

hintjens: I fixed this in zproject...  Did I miss an instance of it?

jimklimov: Most of other changed locations moved variable declaration out of the loop,
e.g. `type varname; for (varname=...`
In the loop for `argn` it was just renamed from `i` but was left declared
inside the `for` clause.

--
Typos courtesy of GMail on my Samsung Android
25.10.2015 19:23 пользователь "Pieter Hintjens" <notifications@github.com>
написал:

hintjens: Good catch. Could you send a pull request? I'm off my pc for a while.

jimklimov: Likewise, on the phone. I'll take a look tomorrow if noone beats me to it.

--
Typos courtesy of GMail on my Samsung Android
25.10.2015 20:23 пользователь "Pieter Hintjens" <notifications@github.com>
написал:

hintjens: OK done it. It's so nice to have a fully working laptop again :-)

Issue #189, opened 2015-05-02
python 3 compatibility


sphaero:
It seems the python binding generation is incompatible with python3

Around https://github.com/zeromq/zproject/blob/master/zproject_bindings_python.gsl#L297
(convenience) methods are used which are deprecated in python 3 because of the newer io system.

from: https://docs.python.org/3.3/c-api/file.html
```
These APIs are a minimal emulation of the Python 2 C API for built-in file objects, which 
used to rely on the buffered I/O (FILE*) support from the C standard library. In Python 3, 
files and streams use the new io module, which defines several layers over the low-level 
unbuffered I/O of the operating system.
```
If you look at https://docs.python.org/2.7/c-api/file.html you can see many methods are gone in python3. 2to3 is not taking this into account therefore the bindings fail in python3.

refering to: https://github.com/zeromq/czmq/issues/998

jemc:
@sphaero - If you or someone else with python knowledge wanted to take a look at improving this, that would be great.  The bindings generator for python was originally written by @twhittock - maybe he/she has some input.

I wonder if it is possible to support both python 2 and python 3 with the same bindings?  If so, finding a way to do that would likely be best.  If not, there would be no problem with adding a second generator that targets python3 (or uses FFI instead, or other procedural differences).

sphaero:
I think it is possible to support both py2 and 3. However the new io systems needs to be used which I think is available in py2 as well. I'll take a look at it but I'd have to admit this is new to me as well. If someone could elaborate on what these methods are doing exactly.

twhittock:
I wrote these bindings. I didn't consider Python3 compatibility when doing so beyond 'obvious' basics. The `*_AsFile` and `*_FromFile` methods are called to provide a sane API to callers. There's no reason the bindings *have* to call these functions, but if they didn't, then the calling Python code would have to work out how to interpret a C FILE* structure and convert it into a Python file itself.

It looks like Python 3 makes this hard to do. I feel this is a general trend with Python 3, and I'm not particularly interested in supporting it, personally.

To fix this, you'll have to use the stuff from https://docs.python.org/3.3/c-api/file.html as well as the stdlib functions like `fdopen` and `fileno` to convert between Python files and C FILE* pointers. The solution probably won't be cross-compatible between Python 2 and 3, and will introduce problems with 2 different systems buffering the data.

twhittock:
[...]

PR #267, opened 2015-10-29
Problem: generated bindings failing on Python 3.

viotti:
Bindings generated for projects that use the buffered I/O (FILE*) C
library calls, e.g. "czmq", will not run on the version 3 of the
interpreter. The actual problem is that the API used to cast files from
Python to C, and vice-versa, is not available on this version. See

  https://docs.python.org/3/c-api/file.html

Solution: update the GSL script to use the Standard C library functions
on Python 3.

### DIFF
--- a/zproject_bindings_python.gsl
+++ b/zproject_bindings_python.gsl
@@ -284,24 +284,60 @@ for project->python_types.type
 endfor
 
 if defined (project->python_types.uses_py_file)
->PyFile_FromFile_close_cb = CFUNCTYPE(c_int, FILE_p)
->PyFile_FromFile = pythonapi.PyFile_FromFile
->PyFile_FromFile.restype = py_object
->PyFile_FromFile.argtypes = [FILE_p,
->                            c_char_p,
->                            c_char_p,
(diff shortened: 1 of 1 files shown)


jemc:
@viotti This will still work on Python 2, right?

By the way, this issue has been discussed before, if you haven't seen this already: https://github.com/zeromq/zproject/issues/189#issuecomment-99558152.

viotti:
@jemc Yes. I did not modify current behaviour for Python 2. I also ran the test battery on both Python 2.7 and Python 3.4.

No, I had not seen #189. I think that issue can be closed, since I've implemented what @twhittock suggested on https://github.com/zeromq/zproject/issues/189#issuecomment-98488634. As for buffering issues, I had to disable it at the standard C library level to make the tests pass. I did not dug for further problems.

BTW, I have a project using Zyre with Python 3 and this version of the `czmq.py` bindings. Everything is working well so far.

jemc:
@viotti - Great, I'll close #189 then!

sphaero:
Nice!!!

PR #270, opened 2015-11-03
minor fix to the generation of python constructors

sphaero:
As mentioned on the ML by Johan Philips the Python constructor can have wrong arguments:

```
from czmq import Zpoller
from zyre import Zyre
node = Zyre("test")
node.start()
poller = Zpoller(node.socket(), None)
Traceback (most recent call last):
   File "<stdin>", line 1, in <module>
   File "/home/jphilips/workspace/czmq/bindings/python/czmq.py [czmq.py]", line 
1782, in __init__
     self._as_parameter_ = lib.zpoller_new(args[0], *args[index()-1:]) # 
Creation of new raw type
NameError: global name 'index' is not defined
```
This PR request removes the undefined index() method.

After merge the bindings of projects might need regeneration.

### DIFF
--- a/zproject_bindings_python.gsl
+++ b/zproject_bindings_python.gsl
@@ -392,7 +392,7 @@ for class where defined (class.api)
         callargs = ""
         for method.argument
             if argument.variadic
-                callargs += "*args[index()-1:]"
+                callargs += "*args[1:]"
             elsif defined (argument.python_coerce)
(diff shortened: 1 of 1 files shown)


twhittock:
Er, this fails when there are more than 1 arguments before the positional
element. I'll make a proper fix now

twhittock:
Please see #271

jemc:
Thanks, @twhittock!

PR #274, opened 2015-11-05
Various improvements

hintjens:


sappo:
@hintjens great cleanup, thanks.

hintjens:
Still need to make it work without a library, only mains.

sappo:
That would be great, the biggest obstacle is probably the dependency to the
czmq prelude.
Am 05.11.2015 7:16 nachm. schrieb "Pieter Hintjens" <
notifications@github.com>:

PR #275, opened 2015-11-06
Problem: can't create dso with versioned symbols

vyskocilm:
Solution: (for autotools)

In case there is src/$(project.libname).sym file, do this
  1.) configure.ac: Add a check if platform linker supports version-script or not
  2.) src/Makemodule.am: append --version-script or --export-symbols to LDFLAGS while
      building a library

Do nothing if sym file does not exists. Tested with GNU linker on Linux, other platforms might need some conditionals ...

### DIFF
--- a/zproject_autoconf.gsl
+++ b/zproject_autoconf.gsl
@@ -386,6 +386,39 @@ AC_ARG_WITH([pkgconfigdir], AS_HELP_STRING([--with-pkgconfigdir=PATH],
     [pkgconfigdir="$withval"], [pkgconfigdir='${libdir}/pkgconfig'])
 AC_SUBST([pkgconfigdir])
 
+
+.if file.exists ("src/$(project.libname).sym")
+# Symbol versioning support: snatched and adapted from libpng:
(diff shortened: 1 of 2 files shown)


sappo:
@vyskocilm I've never seen a *.sym file. What exactly are those and when would the exist?

vyskocilm:
Hi,

basically this brings the ability to add new symbols to the library, while not removing the old ones. The most prominent example is glibs (https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/x86_64/64/libc.abilist;h=c6e3cd42572d7e2a6831e74213ea20f8bb918ccb;hb=HEAD).

Good starting document is libabc's README (https://git.kernel.org/cgit/linux/kernel/git/kay/libabc.git/tree/README#n171).

For our project we do use it as an explicit list of symbols to be exported and not really care about versions.

BTW: I'd say this is not usefull outside of (GNU/) Linux or Solaris world ...

sappo:
okay, thanks

PR #277, opened 2015-11-07
Fixed #258

### LINKED ISSUE #258: zproject requires a class to be defined
Problem: a raw collection of source files could not become a project until a dummy class was added into project.xml even though it is not to be used

hintjens:


hintjens:
You can now make a project file like this:

    <project script = "zproject.gsl">
        <use project = "czmq" />
        <main name = "hello" private = "1" />
    </project>

PR #295, opened 2015-11-18
Improve Ruby bindings

paddor:
* correctly handle variadic arguments
* provide complete list of Ruby keywords


jemc:
Thanks!

paddor:
@jemc You're welcome. :)

PR #305, opened 2015-11-22
enum types in Ruby binding and extended API XML example

paddor:
* add support for enums in the Ruby binding
* fix the processing of `<include ... />`
* GSL `c_method_declaration()`: don't treat all functions as polymorphic
* greatly extend the example API XML `api/myclass.xml`

Examples for the following API elements have been added:

* enums
* the implicit self pointer argument on destructors
* excluded methods (`exclude = "1"`)
* returned pointers which will be owned by the caller (`fresh = "1")
* arguments that are passed by reference (adds one more `*`)
* methods with `va_list` siblings (do_that() and vdo_that())
* arguments of type `byte`
* arguments of type `number`
* arguments of enum type (`enum:myclass.myenum`)
* arguments of type `string`
* arguments of type `buffer`
* arguments of type `size`
* arguments of type `FILE`
* arguments of type `time`
* arguments of type `format`  (which makes the method implicitly variadic)
* explicitly variadic arguments (`variadic = "1"`)
* arguments of type `va_list`
* arguments with overridden C type (`c_type = "foobar_t"`)
* methods that take a polymorphic reference (`polymorphic = "1"`)

Quite a bit of information has been added which could help a developer
of a new language binding.

### DIFF
--- a/api/myclass.xml
+++ b/api/myclass.xml
@@ -4,7 +4,7 @@
     It shows a language binding developer what to expect from the API XML
     files.
 -->
-<class name = "my class" >
+<class name = "myclass" >
     My Feature-Rich Class
 
     <include filename = "license.xml" />
@@ -20,7 +20,7 @@
 
     <!-- Constructor is optional; default one has no arguments -->
     <constructor>
-        Create a new my_class with the given name.
+        Create a new myclass with the given name.
         <argument name = "name" type = "string" />
     </constructor>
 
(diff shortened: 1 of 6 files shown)


hintjens:
I'd be happier with small focussed patches, especially when you are potentially breaking other code (in this case, existing binding generators). Can you please take back this patch and make individual patches for each problem you are solving. Please read the C4. Thanks

paddor:
Of course. I should have worked in different branches to prevent Github from just appending every further commit to the same pull request.

hintjens:
You can work in the same branch and accumulate commits in one pull
request... That is fine. Yet please don't solve multiple issues in a
commit. It makes it too hard for others to understand and critique...
Thanks :)

hintjens:
Oh dear... 

I apologize.

hintjens:
I was using my phone and didn't click through to the pr itself. You did it right, my comments were wrong. So sorry...

paddor:
Oh that's nice to hear. And I was already thinking about how to group the commits into different branches. :joy:

PR #322, opened 2015-11-25
Problem: complicated fresh pointer handling in Ruby bindings

paddor:
Memory of fresh strings isn't free'd automatically when `FFI::Pointer` goes out of scope.

Solution: Wrap it in an `FFI::AutoPointer`.

Consider the following method API:

```xml
  <method name = "encode">
      Encode a stream of bytes into an armoured string.
      <argument name = "data" type = "buffer" />
      <argument name = "data size" type = "size" />
      <return type = "string" fresh = "1" />
  </method>
```

The corresponding C function might look like this:

```c
  char * foo_encode(const byte * data, size_t data_size)
```

Note that the return value is a "fresh" string, which means the caller is responsible for destroying it when finished with it. Also note that it's not a struct of the class defined in this API or any other class. It must be a C style, NULL-terminated string.

It would be nice to have the generated Ruby bindings to automatically release said memory as soon as its corresponding FFI::Pointer goes out of scope.

Luckily, there's FFI::AutoPointer for this. One just has to subclass it and define its `.release` method appropriately. This commit defines it to call libc's `free()`. I believe this is okay, assuming:

* libc is available on every system
* no project-specific function has to be used to `free()` a C style string

Honestly, I'm not 100% sure if this is a good idea. Are my assumptions valid?

Note that:

* this applies only to functions that have `char *` value, which is also marked "fresh"
* automatic release is already done in Ruby bindings for "freesh" class objects (like `zmsg_t` objects) (but using the class' destructor, of course)

### DIFF
--- a/zproject_bindings_ruby.gsl
+++ b/zproject_bindings_ruby.gsl
@@ -86,6 +86,12 @@ function resolve_ruby_container (container)
     elsif my.container.c_type = "const char *"
         my.container.ruby_ffi_type = "string"
         my.container.coerce_to_c = "String($(my.container.ruby_name:))"
+    elsif my.container.c_type = "char *"
+        if my.container.fresh
+            # if it's a fresh string that we have to free() when done with it,
+            # wrap pointer in an FFI::AutoPointer
+            my.container.coerce_to_ruby = "$(project.RubyName:)::FFI::AutoPointer.new($(my.container.ruby_name:))"
+        endif
     elsif my.container.c_type = "byte"
         my.container.ruby_ffi_type = "char"
         my.container.coerce_to_c = "Integer($(my.container.ruby_name:))"
@@ -246,6 +252,20 @@ require_relative 'ffi/version'
 
 module $(project.RubyName:)
   module FFI
+    module LibC
+      extend ::FFI::Library
+      ffi_lib ::FFI::Platform::LIBC
+      attach_function :free, [ :pointer ], :void, blocking: true
+    end
+
+    # Used to wrap fresh pointers so they can be free'd automatically, when the
+    # AutoPointer instance goes out of scope.
+    class AutoPointer < ::FFI::AutoPointer
+      def self.release(ptr)
+        LibC.free(ptr)
+      end
+    end
+
     extend ::FFI::Library
 
     def self.available?


jemc:
@paddor - Sounds right to me - before now we were only handling `fresh` for known zproject-generated classes.  Your assumptions and conclusion are correct, as far as I can tell.

The only thing I don't quite like about this implementation is the extra `AutoPointer` class now in the library namespace - it could possibly be confused with a wrapped class.  I think it could have been done using an `::FFI::Pointer` and the `free` method as an extra argument to the [constructor](http://www.rubydoc.info/github/ffi/ffi/FFI/AutoPointer#initialize-instance_method). That said, this is not a huge problem, so if no one is sufficiently motivated to change it, we can leave it as-is until someone finds it to be enough of a problem to warrant some effort (at the moment, it's not crossed that threshold for me :wink:).

paddor:
@jemc Thanks!

Of course I can change this. PR coming right up!

Asmod4n:
FFI::AutoPointer is roughly 20 times slower than manual managing.

Asmod4n:
Oh, its gotten faster, its only 8 times slower nowadays.

paddor:
@Asmod4n During instantiation or during release? Can we make it faster?

Asmod4n:
@paddor during runtime, you can make it faster by using FFI:: MemoryPointer.from_string(fresh_string) and then free it.

Asmod4n:
Or if you want to get a ruby string directly, something like this
```ruby
module FFI::CZMQ::Helpers
  def self.fresh_string(cstring)
    rstring = cstring.read_string
    LibC.free(cstring)
    rstring
  end
end
```

Asmod4n:
@paddor during instantiation, FFI::AutoPointer uses ObjectSpace.define_finalizer, which is super slow.

paddor:
Yeah that sure would work.

I don't know, but doesn't a good binding offer the possibility to keep working with the C string, possibly passing it on to another function? Or do you think that's an edge case and we just require any returned C string which someone would like to pass to another C function to have to round-trip through a Ruby string?

Asmod4n:
When i was writing my ffi wrapper for libsodium i first allocated all buffers via ffi so it was "closer" to the c code and only at the end returned the ruby string, it turned out to be faster just to do everything in ruby, which was a bit surprising.

Asmod4n:
It even was faster to pass around the ruby buffers then the ffi buffers to other c functions. But this was roughly a year ago, so things might have changed.

jemc:
Since it's a "fresh" string where ownership is transferred to the caller, and the caller is Ruby, it's probably okay to reallocate it as a Ruby string, especially if it's actually more performant in common cases as @Asmod4n claims.

If it were not a fresh string, then we wouldn't want to swap it out for a Ruby string.

paddor:
Okay, makes sense. I'll change it to that, then. And benchmark it, out of curiosity.

jemc:
I'd imagine the size of the string in question makes a big difference in the benchmark.  For very large strings, I'd expect to pay a big penalty for the allocate-and-copy to Ruby.

paddor:
Just benchmarked it. Using the finalizer, it seems to be ~2x slower for short strings, and only slightly (~15%) slower with long strings.

```
$ ./benchmark_fresh_cstring.rb
SHORT STRING (20 bytes)
==============================================================================
Calculating -------------------------------------
    finalizer free()    12.508k i/100ms
    immediate free()    30.465k i/100ms
-------------------------------------------------
    finalizer free()    173.257k (±48.7%) i/s -    612.892k
    immediate free()    370.366k (±10.0%) i/s -      1.858M

Comparison:
    immediate free():   370366.5 i/s
    finalizer free():   173256.8 i/s - 2.14x slower


LONG STRING (2000 bytes)
==============================================================================
Calculating -------------------------------------
    finalizer free()     4.500k i/100ms
    immediate free()     5.332k i/100ms
-------------------------------------------------
    finalizer free()     55.104k (±22.5%) i/s -    252.000k
    immediate free()     62.283k (± 8.6%) i/s -    309.256k

Comparison:
    immediate free():    62283.3 i/s
    finalizer free():    55103.5 i/s - 1.13x slower

./benchmark_fresh_cstring.rb  26.23s user 2.25s system 99% cpu 28.584 total
```

I didn't disable the GC for these tests, though. See this [gist](https://gist.github.com/paddor/760caa83c6ebd2eb20ab).

How did you benchmark it?

PR #323, opened 2015-11-25
Problem: superfluous class in Ruby binding

paddor:
Solution: Get rid of it and pass the relevant functionality as a Method
object (which is the preferred way according to FFI documentation
anyway).

As discussed in #322.


jemc:
Thanks!

PR #327, opened 2015-11-26
Cleanups

hintjens:


paddor:
@hintjens Looks great, thank you. However, for some reason, the Ruby bindings generator skips the included zsock_option methods (even though they're actually included by `resolve_includes` (I verified) and at least get resolved in `resolve_ruby_method` (I verified)). But later in the `.for method` loop, they just get skipped and I have no idea why.

In case you know something off the top of your head, please let me know.

Otherwise I'll keep looking...

hintjens:
No idea off the top of my head, I'll take a look though.

paddor:
Thank you. (I'll be AFK for the next 3h.)

paddor:
False alarm! It used to generate FFI-stuff for the zsock_option functions twice! Now it's correct. But of course the diff looked like it got removed. :+1:

PR #333, opened 2015-11-27
Problem: implied new/destroy/print are surprising

### LINKED ISSUE #332: Problem: implied methods are too surprising
Adding new, destroy, print *unless* these are explicitly specified with 'exclude' is nasty.

If we did want to exclude these methods we'd use an element 'exclude' rather than a method with a property.

However even this still makes a simple example more complex than it needs to be. 

Solution: add 'test' by default since this is mandatory for CLASS. The other methods do not need any special treatment. The 'exclude' attribute serves no purpose.

hintjens:
Solution: remove this logic. Classes that need these methods should
simply define them. The extra cost is close to zero and it gives a
what-you-see-is-what-you-get API definition.

If we absolutely need to hide details that we see often in classes
there are other techniques. The best for classes (as it is well
known and not surprising) is inheritence. It is easy to do with
GSL. We've done this before in class models. The downside is you
start to get too many layers. The upside is you can ensure all
classes of a certain type are structured the same way.

At this stage I'd much prefer simplicity over abstraction. We are
already rather deep.

Fixes #332

### DIFF
--- a/zproject_class_api.gsl
+++ b/zproject_class_api.gsl
@@ -199,41 +199,21 @@ function resolve_c_method (method, default_description)
     endfor
 endfunction
 
-# Resolve missing or implicit details in a C class model
+#   Resolve missing or implicit details in a C class model
 #
-# Here, "class" refers to a <class/> entity in the model XML.
+#   Here, "class" refers to a <class/> entity in the model XML.
 #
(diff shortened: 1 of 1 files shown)


hintjens:
I'm aware this breaks CZMQ and other projects. Better some pain now than this confusion for the next years.

PR #334, opened 2015-11-27
Problem: API examples describe removed exclude attribute

paddor:
Solution: remove those examples

In addition to #332


hintjens:
Thanks.

PR #340, opened 2015-11-30
Problem: mutable property had wrong default

hintjens:
Solution: default silently to 1 for types other than strings/buffers.


paddor:
Even though I'm using the latest version of zproject, when I run `gsl project.xml` in CZMQ, it removes all `const ` qualifiers from all parameters and return types in the header files.

Can you check this on your side?

paddor:
Scratch that. Now it's fine.

PR #346, opened 2015-12-01
Revert "PROBLEM: "include/$(project.header:)" generates a wrong file name"

sappo:
Reverts zeromq/zproject#345

This is an invalid problem. CLASS does not allow a minus in file names instead everything MUST be separated by an underscore.


sappo:
@karolhrdina please consider changing your project's name to `alerts_list` or `"alerts list` in your project.xml

PR #368, opened 2015-12-07
Convert excess constructors into singleton methods

hintjens:


paddor:
Just ran it locally. It breaks at least the Ruby binding (it worked fine with multiple constructors before). The problem is that these new "methods" aren't marked as fresh, I think.

paddor:
Actually I don't think this is elegant. To work around this, I have to turn this GSL

```
.for constructor
      # CTOR DEFINITION
.endfor
.for destructor
      # DTOR DEFINITION
.endfor
.for method
      # METHOD DEFINITION
.endfor
```
into:

```
.for constructor
      # CTOR DEFINITION
.endfor
.for method where method.is_constructor = 1
      # CTOR DEFINITION (AGAIN)
.endfor
.for destructor
      # DTOR DEFINITION
.endfor
.for method where method.is_constructor = 0
      # METHOD DEFINITION
.endfor
```

Isn't there another way to solve this?

jemc:
I would be in favor of reverting 3d9bdde.

It shouldn't be a big deal for the other bindings to support multiple constructors.  As @paddor mentioned, in some cases this is actually less convenient.

See also https://github.com/zeromq/czmq/pull/1220#issuecomment-162694411 - the only reason the python bindings don't support them is because @twhittock  didn't know about them when writing the generators, not because it is a significant burden.

sappo:
The problem is invalid for the qml and qt bindings as the both support multiple constructors

paddor:
@sappo I don't get it. Then why did you merge?

sappo:
The first problem @hintjens also pushed is very valid. The other we can quickly revert.

paddor:
:+1:

hintjens:
Yeah, sorry about that. Seemed worth trying as a quick fix.

PR #373, opened 2015-12-08
Problem: installation of tmpfiles.d files is hacky in zproject

vyskocilm:
Solution: add support for main attribute tmpfiles = "1" (like service =
"1"), which generates template of proper tmpfiles.d config snippet.

Supported in autotools and spec generators

### DIFF
--- a/zproject_autoconf.gsl
+++ b/zproject_autoconf.gsl
@@ -442,6 +442,9 @@ AC_CONFIG_FILES([Makefile
 .   for project.service
                  src/$(service.name).service
 .   endfor
+.   for project.main where main.tmpfiles ?= 1
+                 src/$(main.name).conf
+.   endfor
                  ])
 .else
 AC_CONFIG_FILES([Makefile])
--- a/zproject_automake.gsl
+++ b/zproject_automake.gsl
@@ -205,6 +205,10 @@ src_$(main.name:c)_service_DATA = src/$(main.name).service
 src_$(main.name:c)_service_DATA = src/$(main.name)@.service
 .   endif
 .   endif
+.   if main.tmpfiles ?= 1
+src_$(main.name:c)_tmpfilesdir = @prefix@/lib/tmpfiles.d/
+src_$(main.name:c)_tmpfiles_DATA = src/$(main.name).conf
+.   endif
 endif
 .endfor
 .for bin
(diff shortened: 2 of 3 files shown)


hintjens:
I'll merge this yet I'm not sure it's well formulated. tmpfiles is a concept specific to these generators, and should not be visible in the project model. I don't (as a user) have a clue what this means or why I should use it. Can we abstract this somehow?

hintjens:
Same goes for main.services, which is undocumented... @miska, we need to be careful about randomly extending the model language with low level attributes.

vyskocilm:
I completelly agree - however as long as this is not documented, it seems to be fine.

Right now from reading the code
no_config="1" - suppress generation of systemd service file
service="1" - create systemd service file
services="1" - create systemd service file template
tmpfiles="1" - create a recipe for systemd-tmpfiles, which is systemd component to manage volatile directories

Previously I have installed tmpfiles.d config files from Makemodule-local.am, but this is not accessible for spec file generator.

vyskocilm:
Ad I completelly agree - however as long as this is not documented, it seems to be fine to experiment first.

Technically all of those are extra files to be installed into different well-known locations, so maybe there is better model for them.

vyskocilm:
More I am thinking about the case - I'd say we can workaround it by putting those files to Makemodule-local.am and change spec generator that way that filelist will be generated from filelist.

This way we can get rid of most of extra attributes, while bypassing the model.

vyskocilm:
I made PR#374 to revert this change.

I have prototyped different approach, which sounds mode model-like in my mind.
@hintjens  please review
```
    <main name = "foo" service="1"> Some daemon
        <extra type="systemd-tmpfiles" />
    </main>
```

The idea is that type will contain the name and installation location implicitly - I wrote simple resolving function to zproject_projects.gsl, which now becomes only one source of "wisdom" and other script are simply looping through main->extra and pringint the stuff.

With this approach one can add type "systemd-service" and "systemd-service@" (or "systemd-service-template"??) and deprecate all attributes for main element.

I am going to create PR implementing such model for autotools and spec file.

hintjens:
I like this more; you could use <option target = "systemd"> perhaps instead
of <extra>. It is a decent choice to have space for low-level customization.

Still, when designing a model (and this isn't easy), imagine that you want
to target an arbitrary number of backends. So, services on Linux, services
on Windows. Using the same model. As we do for the API model, and language
bindings.

We can do this one step at a time. What I like about a sub-element is that
it clearly separates non-portable details from the main model.

vyskocilm:
For `option` or `extra` I have no strong prefference. Right now all of our project uses Makemodules-local.am and manually changed spec file which works well. So it is fine to experiment with this and document the result.

To arbitrary number of backends - I am pretty sure that every simple build model (cargo, python setuptools, ...) started to be nasty and complex when it comes to installation and integration with target systems ;-)

Basically I see two kinds of stuff we're interested in - one is very technology/platform specific like `systemd-tmpfiles`, which most likely does not have an equivalent in other systems. The second one are cross-platform concepts like service (or maybe firewall recipes to open the dedicated ports by default) you're probably more interested in.

I was thinking to model cross-platform concepts as subelements too, however this can be revised later. For next steps I do recommend to
 1.) remove `services` (ending with s) and adapt the code that main.name.service or main.name@service.in will both match
 2.) remove `no_config` and add `config="1"` to actually opt into default config file
 3.) add flag similar to `script="1"` and deprecate the bin service elements, so all those extra platform integration will get available for added scripts as well.

With all those zproject will become close to perfect for all of use cases I found with quite nice and cross-platform model. I'll send pull requests for those later.

hintjens:
Target systems are nasty, usually we try to hide this with smart defaults,
without giving the user options except when absolutely necessary. Or, make
options a run-time thing, e.g. switches passed to command line scripts.

PR #394, opened 2015-12-16
Fixing VS20xx generation

hintjens:


hintjens:
@zeromq/zproject-maintainers if you are building on Windows, please review this change.

I've killed the duplicate MSVC backends and started generating more of the necessary packaging, like .gitignore, build.bat, the .sln file. I've also removed all MSVC-generated files from CZMQ and Zyre, so probably I broke something.

As a rule, please *do not add files to the build directories* ever. If you need to add a hand-built file, or modify .gitignore, do it in zproject.

PR #396, opened 2015-12-18
Solution: auto-underscore python method names

c-rack:


hintjens:
It's easier to parse the PR if you use the same Problem: from the commit.

c-rack:
Sure!

PR #399, opened 2015-12-19
Problem: build scripts don't need full git history

c-rack:
Solution: add `--depth 1` parameter to `git clone` commands


hintjens:
Nice. Problem: checking out full history for Ci builds is slow, perhaps.

PR #436, opened 2016-01-08
Problem: unecessary constant attribute printed

vyskocilm:
Solution: skip it for string and buffer arguments, which are imutable
by default


vyskocilm:
need to update

PR #450, opened 2016-01-13
Problem: can't generate zclock API model

vyskocilm:
Solution:
 * fix comment aligment
 * remove unamed arguments
 * add clock type

### DIFF
--- a/zproject_class_api.gsl
+++ b/zproject_class_api.gsl
@@ -100,6 +100,8 @@ function resolve_c_container (container)
         my.c_type = "off_t"
     elsif my.type = "time"
         my.c_type = "time_t"
+    elsif my.type = "clock"
+        my.c_type = "int64_t"
     elsif my.type = "FILE"
         my.c_type = "FILE"
         my.stars = "*"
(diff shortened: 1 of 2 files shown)


zproject_class_api.gsl
```diff
@@ -100,6 +100,8 @@ function resolve_c_container (container)
         my.c_type = "off_t"
     elsif my.type = "time"
         my.c_type = "time_t"
+    elsif my.type = "clock"
```
jemc: I think the naming of this type is a bit dubious.

What does everyone think of `signed/8` instead?

hintjens: I'd prefer something more abstract, rather than the actual type (in some
cases it's useful to say, "64 bits"... not here). Maybe "msecs"?

vyskocilm: As @hintjens said on zeromq-dev, the real problem is unfortunate API of zclock which can't be expressed by API model. The naming would be confusing all the time imho.

But this is C4.1 project, so lets solve it by pull requests :-)

PR #475, opened 2016-01-22
Problem: can't have more extra files for one main

vyskocilm:
Solution:
 * correctly iterate through all extra's of main
 * add attribute cond (condition) to distinguish files belong to
   conditional installation or others
 * fix redhat and debian packaging as well

### DIFF
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -502,6 +502,11 @@ AC_CONFIG_FILES([Makefile
                  doc/Makefile
                  src/$(project.libname).pc
 .   endif
+.   for project.main where defined (main->extra)
+.       for extra where extra.cond ?<> "with_systemd"
+                 src/$(extra.name)
+.       endfor
+.   endfor
                  ])
 .else
 AC_CONFIG_FILES([Makefile])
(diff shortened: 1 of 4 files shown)


hintjens:
I've still some questions about the need to specify target-specific
properties. What are the actual semantics of saying "with_systemd" as
compared to not saying it? Do you have a mix of files that get passed to
systemd, and do not? What does this mean?

I'd like to always find ways to abstract out these details.

vyskocilm:
Hi,
I currently (miss) use extra to install additional files along with daemon. But idea behind cond="with_systemd" is that's *internal* marked to simplify generation of systemd-related integration (timers, path units, ...) in the future. My idea is to simplify the service="1", services="1", no_config="1" in main's model sometime ...

BTW: the more I am thinking about it the more I am convinced that Makemodule-local.am + some way to influence the generated list installed files for packages might be much easier and simpler solution ...

hintjens:
Yes, we could have a hook to a hand-made Makemodule file that does the
custom install. Remember that every syntax in zproject needs eventually to
be documented and explained to others.

vyskocilm:
Hi,

Ok, I'll then try and enhance local hooks and make PR to remove this stuff. It's probably to complex to be modeled properly.

PR #476, opened 2016-01-22
Problem: zproject_python output is broken if a class or method descri…

asokoloski:
…ption ends with a double quote character

Solution: Put docstring quotes on their own lines.
This conforms to PEP-0257 (Docstring Conventions) as well.


asokoloski:
Ta!

PR #488, opened 2016-01-25
Two patches

asokoloski:
The zproto headers one is the slightly more interesting one.

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -412,6 +412,13 @@ $(c_actor_declaration (actor):)
 $(c_method_declaration (method):)
 
 .   endfor
+.-
+.   for class.variable where draft = my.draft
+//  $(variable.description:no,block)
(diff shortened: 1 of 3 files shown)


asokoloski:
Thank you!

sappo:
@asokoloski I'm note sure weather your first problem is valid. I agree that zproject messes up the zproto header. But you can restore that with `make code`. Thereby the API file is only effectively used by the bindings.

asokoloski:
Hmm, ok.  I'm still a bit new to all this stuff, sorry.

This reasoning behind this is mainly coming from here:
https://github.com/zeromq/malamute/issues/115

But I'm not too bothered.  make code seems fine by me.

asokoloski:
I'll push a revert.

asokoloski:
I mean, revert and make a pull request :)

sappo:
The zproject zproto integration currently doesn't work well. We should fix that!

PR #489, opened 2016-01-25
Problem: Previous patch adding variables was not really necessary.

asokoloski:
...as you can re-generate headers pretty easily with "make code".

Revert "Problem: Re-generating zproto headers with zproject API model drops global verbose flag"

This reverts commit 24945ab93ecee3df2c1c878ed3f6e5865204072b.


hintjens:
In any case I need to remove the global variable as zproject does not
support these (and they're nasty). Better is a singleton that modifies a
hidden static variable.

PR #493, opened 2016-01-26
Problem: Every build target shares the same project model tree.

asokoloski:
This makes name collisions likely.

Solution: Make a fresh copy of the project tree before building each target.
This requires slight changes to the java and java_msvc targets, which depended
on shared mutation of the model and order of the build targets.

Used czmq to test that output did not change.


hintjens:
Neat ☺

asokoloski:
:)

PR #500, opened 2016-01-28
Problem: Python CFFI target broke the Python target

hintjens:
It was using functions with the same name, however we're using
a global scope, so this broke the python target.

Solution: use unique function names.

A better solution would be to wrap each target in its own scope.
I'll experiment with this.


hintjens:
@zeromq/zproject-maintainers would you merge this?

PR #511, opened 2016-01-29
Problem: python bindings are not easy to install

asokoloski:
Solution: generate a setup.py script.

To install bindings, run "python setup.py install".


asokoloski:
Thank you!

PR #522, opened 2016-02-02
Problem: `--enable-drafts` is without effect

c-rack:
Solution: use `$enableval` instead of `$withval`


hintjens:
Thanks for catching this. I'd seen it and then forgot to make the change.

c-rack:
I have not worked with autotools that much; it was a good opportunity to learn more about it :-)

PR #539, opened 2016-02-10
Problem: the installation model is uber messy

vyskocilm:
Solution: get rid of it

This patch effectivelly drops
 * services and no_config for main
 * project.services
 * main.extra

It preserve service attribute for main with the same semantics like back
then (one service file and one config file). Installation is still behind
WITH_SYSTEMD_UNITS variable, which seems like the most usefull idea of
all the changes we did.


vyskocilm:
cc @hintjens  - you will love this patch ;-)

hintjens:
I like stuff I don't understand disappearing :-)

vyskocilm:
Cowboy programming did not end well in this case

hintjens:
It's always good to experiment... And designing good models takes time.

PR #541, opened 2016-02-11
Problem: debian packaging broken after adding uuid dependency

vyskocilm:
Solution: resolve debian_name and spec_name from known use and add
proper debian_name for uuid


vyskocilm:
BTW: this will open a possibility to simplify the dependency generator for redhat and debian packaging - names can be resolved before.

PR #552, opened 2016-02-19
Problem: my.topdir variable is unkown

sappo:
Solution: use project.topdir which seems to be the correct one


hintjens:
Thanks for doing this, I'd also fixed it and not pushed yet.

sappo:
@hintjens You're welcome. Do you have an explanation for #553 which still fails for me.

PR #557, opened 2016-02-19
Travis fixes

sappo:
* Allow building dependencies from tarballs
* Execute buildconf instead of autogen.sh which is used e.g. by libcurl


bluca:
Awesome, thanks for implementing this! :+1:

PR #559, opened 2016-02-22
Problem: drafts are confusing on many projects

### LINKED ISSUE #550: Problem: drafts are confusing
We often get tripped up by projects with no stable API, where 'nothing works'. In Zyre we fixed this making the API stable. Another workaround is to define ZYRE_BUILD_DRAFT_API. It's not elegant. 

Solution: if project has only draft classes, treat as --enable-drafts

hintjens:
It's clumsy that projects that don't even care about this aspect
are forced to use it, or they don't build properly. That is, the
project header is useless on projects consisting only of draft
classes, which is the default for new projects.

Solution: always build drafts on projects that have no stable
public classes.

Fixes #550


bluca:
:+1:

PR #571, opened 2016-03-09
remove test code and fix comment

gregoryyoung:
The generated code was generating tests that lock up selftest by default.

I am not sure yet what they should generate but probably not what they are.

### DIFF
--- a/zproject_skeletons.gsl
+++ b/zproject_skeletons.gsl
@@ -219,7 +219,7 @@ $(actor.name:c)_start ($(actor.name:c)_t *self)
 }
 
 
-//  Stop this actor. Return a value greater or equal to zero if initialization
+//  Stop this actor. Return a value greater or equal to zero if stopping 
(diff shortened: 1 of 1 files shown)


hintjens:
Nice catch. You'll want to add your own test steps depending on your actor. The generated skeleton should probably just create, then destroy the actor (which tests its handshaking and nothing more).

gregoryyoung:
I'll update as that then.

On Thu, Mar 10, 2016 at 12:19 AM, Pieter Hintjens <notifications@github.com>
wrote:

> Nice catch. You'll want to add your own test steps depending on your
> actor. The generated skeleton should probably just create, then destroy the
> actor (which tests its handshaking and nothing more).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zeromq/zproject/pull/571#issuecomment-194535617>.
>



-- 
Studying for the Turing test

PR #579, opened 2016-03-21
add notests to zproject

gregoryyoung:
Problem: don't want self tests generated for some thing (bringing in a class that is generated by protobuf-c is a good example)
Solution: add notest to class/actor which will optionally not generate self test and will not include file in self_test

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,8 +245,10 @@ zproject's `project.xml` contains an extensive description of the available conf
         Classes, if the class header or source file doesn't exist, this will
         generate a skeletons for them.
         use private = "1" for internal classes
+        use notest = "1" if you do not want self test generated
     <class name = "myclass">Public class description</class>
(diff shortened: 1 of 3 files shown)


hintjens:
Not sure about negative names,  will think of another phrasing.

gregoryyoung:
I couldn't come up with a positive name to describe the lack of the
functionality :) Can't make it "with_tests" etc as that would break
existing working project.xml

On Mon, Mar 21, 2016 at 2:18 PM, Pieter Hintjens <notifications@github.com>
wrote:

> Not sure about negative names, will think of another phrasing.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/zeromq/zproject/pull/579#issuecomment-199306359>
>



-- 
Studying for the Turing test

hintjens:
OK, see #580

PR #580, opened 2016-03-24
Problem: 'notest' is a weak name for a property

hintjens:
In general negative names are difficult for users (they lead to
double negatives like 'notests is 0 by default').

Solution: use 'selftest' which defaults to 1 on actors and classes

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,45 +7,45 @@
 
 
 **<a href="#toc2-11">Overview</a>**
-*  <a href="#toc3-18">Scope and Goals</a>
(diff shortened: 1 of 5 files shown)


bluca:
Looks like CI isn't happy:

gsl/4 I: Processing project.xml...
(zproject_class.gsl 190) Undefined expression: selftest

Want to fix it before merge, or want me to merge it anyway?

hintjens:
OK, thanks. Made #581

bluca:
:+1:

PR #587, opened 2016-04-04
Problem: Windows files do not have zproject headers

opedroso:
Solution: Various per below details.

* First of all, these changes were tested by using zproject/tstgenbld.sh:
```sh
```

After the projects czmq, malamute and zyre were regenerated, built, and make check was run on them, making sure that no new build problems or bugs were introduced.
Same was repeated in the Windows platform once Linux verification was complete.

* Added Windows batch scripts a zproject header

```
project.generated_warning_header ?= "\
project.generated_warning_header_for_dot_bat ?= "\
@::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
@::  THIS FILE IS 100% GENERATED BY ZPROJECT; DO NOT EDIT EXCEPT EXPERIMENTALLY  ::
@::  Read the zproject/README.md for information about making permanent changes. ::
@::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::"
```

* Made every Windows related files be CR+LF terminated as opposed to LF only termination. [various files]
```
.# change terminator for Windows file (CR+LF)
.terminator="\r\n"
.output "$(my.path)/build.bat"
@ECHO OFF
$(project.generated_warning_header_for_dot_bat:)
```
* Always restored it back when done with this macro or file
```
.terminator="\n"
```

* Added a .gitattributes file that tells get to check them in a the file stands instead of doing automated conversions during checkin. [zproject\zproject_git.gsl]
```
if !file.exists (".gitattributes")
    echo "Generating initial .gitattributes file"
    output ".gitattributes"
    ># disables auto CRLF conversion for all files; create the file correctly and it will be allright
    >* -text
endif
```

* Added support for "clean" argument to build.bat script:
```
.output "$(my.path)/build.bat"
@ECHO OFF
$(project.generated_warning_header_for_dot_bat:)
:: Usage:     build.bat [Clean]
@setlocal
```

* Enabled generation of PDBs for release targets. [zproject\zproject_vs20xx_props.gsl]
```
      ~~<!--~~<DebugInformationFormat>ProgramDatabase</DebugInformationFormat>~~-->~~
      ~~<!--~~<GenerateDebugInformation>true</GenerateDebugInformation>~~-->~~
```

* There also some fixes to Windows batch scripts that did not work during error conditions, such as using a goto END but not END label in the script.

* Learned that after you build projects using Linux one must not build in Windows using the same disk files, since  that will break the Windows build.


opedroso:
Hi Pieter,

I have a question on the followup of this PR.
Now that the ZPROJECT changes have been merged, I would like to check in
the generated files for CZMQ, MALAMUTE and ZYRE.

These would take advantage of the fixes to be batch scripts, plus the new
headers.

What is the process? Should I go ahead and do these PRs or should I wait
for somebody else to regenerate and PR them? If somebody else did it, it
would add a check and balance that the build also works for them (which I
have no doubt since I have done them but there is always the possibility of
a "works on my machine" kind of thing).

Another question is how to approach the newly created files. An example
would be the non existent vs2010, vs2012 and vs2013 Windows build files
that are not in malamute/builds/msvc but do get created by zproject.
I am currently using vs2013, so I know this one builds without problems,
but I will not build using vs2010 and vs2012.
Should I check them in?

Still talking malamute, ZPROJECT generates these but these directories do
not currently exist in github.

   - What about the bindings for nodejs, qmi and qt?

   - What about the builds subdirectories for android, cygwin, mingw32?


   - Same question applies to packaging directory, which has subdirs for
   debian and redhat

In general, do the same answers apply to others projects as well?

In this picture, the right side are the same projects after they were
regenerated, on the left, as they current are in github:
[image: ShouldIcheckThemIn.png]

Thanks for your guidance,
Osiris

sappo:
@opedroso it's great that you take so much care not to do any harm. The current process which IMO works best is just to generate everything and fix things as they occur. Sure this approach might break master but in experience fixes are applied really fast if any. So feel free to open these PRs.

bluca:
@sappo @hintjens @opedroso - tstgenbld.sh regenerates overriding the project.xml target parameter:

```
cd ${GITPROJECTS}/$project &&
${BUILD_PREFIX}/bin/gsl -target:* project.xml &&
```

This is fine for testing purposes, but is it the right thing to check the result of that in? Shouldn't we respect the settings of project.xml and check in git only the result of `gsl project.xml` instead?

sappo:
@bluca good catch. `tstgenbld.sh` should definitely respect the `project.xml`'s target setting otherwise comparing the results as @opedroso suggest in the README would show wrong results.

opedroso:
What should the correct command line be for it then?

I got that example (only one I found) from zproject/ci_build.sh.


( cd czmq && PATH=$PATH:${BUILD_PREFIX}/bin gsl -target:* project.xml
) || exit 1

Should it be run like this instead?

${BUILD_PREFIX}/bin/gsl project.xml

?
​

bluca:
Yes, without the `-target` parameter. The zproject's CI uses it as it's useful to test all targets.

opedroso:
You should revert the changes to Zyre, Malamute and czmq before I can
regenerate them without using the "-targets:*' argument.

But I won't have time to do it today.

I will be driving 4 hrs to Atlanta airport to pickup my daughter coming
back from an Europe tour. Then 4 hours back trip...

hintjens:
No worries, send new commits when you can ☺

opedroso:
I am worried that not so many files would be checked in though if not using
-targets:*

It may make sense to pull it for now.

bluca:
There's no rush, don't worry :-)
When you have time, I'd say the quickest thing would be to git revert the previous commit and then regenerate.

PR #592, opened 2016-04-11
- preparation for out of source build 

ostri:
- moved from insource builds to out of source builds 
- instructions how to create an eclipse project on top of malamute project

### DIFF
--- a/zproject_cmake.gsl
+++ b/zproject_cmake.gsl
@@ -15,7 +15,29 @@ register_target ("cmake", "CMake build system")
 .macro target_cmake
 .output "CMakeLists.txt"
 $(project.GENERATED_WARNING_HEADER:)
-
+#-----------------------------------------------------------------------
+#
+# in order to build out of sorce with cmake perform the following actions:
+# 1. in parallel with top project directory create build directory
+#    mkdir build_<project name>
+#    e.g. mkdir build_malamute
+# 2. move to the build directory
+#    cd build_malamaute
+# 3. establish two soft links
+#    - to top project directory  ln -s ../malamute top
+#    - to source directory       ln -s ../malamute/src src
+# 4. generate makefiles from cmake script(s)
+#    cmake ../malamute
+# 5. build the project
+#    make
+#
+# if you wish to work with eclipse
+# - select file/import/from existing sources
+# - select build build directory and confirm
+# - new project will be generated and populated with the source files (e.g. malamute)   
+#    
+#
+#
 ########################################################################
 # Project setup
 ########################################################################
@@ -247,10 +269,10 @@ target_link_libraries(
 .endfor
     ${OPTIONAL_LIBRARIES}
 )
-set_target_properties(
-    $(main.name)
-    PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${SOURCE_DIR}/src"
-)
+#set_target_properties(
+#    $(main.name)
+#    PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${SOURCE_DIR}/src"
+#)
 .endfor
 
 ########################################################################


hintjens:
Before I merge this, I'm going to ask what's broken about out of source builds now? Is is the link error?

```
malamute.c:(.text+0x417): undefined reference to `mlm_server'
```

Please do be explicit about the Problem: for every commit, and explain how your solution fixes the problem. See the C4 guidelines if you're unclear on this.

hintjens:
@ostri can you answer my questions? thanks.

ostri:
Pieter,

Yesterday I was out of country, and I was not able to answer.

On 11.4.2016 13:29, Pieter Hintjens wrote:
>
> Before I merge this, I'm going to ask what's broken about out of 
> source builds now? Is is the link error?
>
> |malamute.c:(.text+0x417): undefined reference to `mlm_server'|
No I was not chasing this bug. With me it compiles and links without the 
problem.

> Please do be explicit about the Problem: for every commit, and explain 
> how your solution fixes the problem. See the C4 guidelines if you're 
> unclear on this. 
*Informal explanation:*
Malamute has two build systems: autotools and CMake. CMake way mimic the 
autotools way (e.g. in-source builds). Even more, standard features 
(again out of source build) which comes at no cost with cmake and with 
high cost with make, are bypassed so that in-source building is possible.
Don't know real reason why there are two build systems. I assumed that 
is because you decided to gradually replace old way of building the 
project with the new one. cmake is clearly on higher abstraction level 
that autotools, although they do not 100% overlap.

Pushing cmake build in the direction of real build (currently it 
performs only compile/link and installs). Final goal is of course being 
able to build everything through cmake (i.e. touch project.xml and all 
sources are regenerated and recompiled).

My "patch" does not break anything for back (just go to malamute folder 
and do "cmake . && make" and check by yourself), but it opens the closed 
way to implement overall build via cmake in a "proper cmake way".

If you decide that this is explanation is not good enough to confirm the 
patch its fine with me. I'll keep it in my clone of the project. But now 
I'd like to concentrate on real issues of mlm:
- add curve support (I have it already locally, but I won't go into 
changes unless I finish succesfuly the trivial one, like you suggested)
- add retry support for the services
- add multibroker support (i.e. simple cluster)

*Formal explanation:*


      Patch Requirements

  * Maintainers and Contributors MUST have a Platform account and SHOULD
    use their real names or a well-known alias.

Done

  * A patch SHOULD be a minimal and accurate answer to exactly one
    identified and agreed problem.

Done - "preparation for out of the source build"

  * A patch MUST adhere to the code style guidelines of the project if
    these are defined.

Done - (just few comment characters :-) )

  * A patch MUST adhere to the "Evolution of Public Contracts"
    guidelines defined below.

Done

  * A patch SHALL NOT include non-trivial code from other projects
    unless the Contributor is the original author of that code.

OK.

  * A patch MUST compile cleanly and pass project self-tests on at least
    the principle target platform.

Passes on fedora 23 (BTW what is principle target platform?)

  * A patch commit message SHOULD consist of a single short (less than
    50 character) line summarizing the change, optionally followed by a
    blank line and then a more thorough description.

I believe commenting out few lines does not require long explanation, 
but I can do it if it is required. ;-)

  * A "Correct Patch" is one that satisfies the above requirements.

hintjens:
Let me discuss this patch a little more first.

* We have learned not to try to "standardize" on a single build system. For instance CMake cannot target embedded systems, and Autotools cannot target Windows.

* The CMake build system *already* allows out of tree builds, and in fact forces it. So I don't see what the reason is for the change (this is my real question).

hintjens:
If you like I'm on the #zeromq IRC channel ATM.

ostri:
> If you like I'm on the #zeromq IRC channel ATM.

We mised. ;-)

> We have learned not to try to "standardize" on a single build system. For instance CMake cannot target embedded systems, and Autotools cannot target Windows.

point accepted; several build systems allowed

> The CMake build system already allows out of tree builds, and in fact forces it. So I don't see what the reason is for the change (this is my real question).

I must politely disagree with you. Just check where binaries and libraries are after build with cmake.

```
set_target_properties(
    malamute
    PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${SOURCE_DIR}/src"
)
```

hintjens:
OK, this is where you need to be really clear when you state a problem. Out
of source builds work, but results are placed into /src.

Problem: CMake build results are (wrongly) put into src dir

Solution: put these ???

And then I can argue properly, since putting the build results into src is
a deliberate choice so that the CMake build produces the same results as a
default autotools build.

ostri:
Hi Pieter,

We are running in cycles. ;-)

My mail "Wed, 13 Apr 2016 08:31:49 -0700" to you

*Informal explanation:*
Malamute has two build systems: autotools and CMake. CMake way mimic the 
autotools way (e.g. in-source builds).

So you want to stick with in-source builds. No problem just say so. 
Since I do not want to start holly war in-source vs out--of-source I am 
dropping my request for patch, like I already said in my mail (cited above).

Best regards

Matjaž

hintjens:
No holy war, it's just very valuable to be explicit about the problems we
want to solve, because other people may disagree with the problem itself.
The solutions are secondary then. :)

I don't mind changing this, if it makes things better. We should ask
whether other build systems (autotools, gyp) need the same changes, and
whether we want to create a standard place for libraries and executables.
(Windows builds have the same issue.)

Consistency is always worth it, if we don't break existing agreements.

sappo:
I'm closing this as the PR is in violoation to contribution guideline namely it does not solve one clear and agreed problem.

PR #603, opened 2016-04-30
Please merge now that Travis build succeeds. ;-) Travis does not check if zproject is up to date with itself

gotcha:
Problem Travis does not check if zproject is up to date with itself
Solution:
Add check to Travis


bluca:
The issue is that the default test does not make check zproject, but tries to regen czmq, so it's a bit of a special case :-/

gotcha:
@bluca Right, I was fixing it.

PR #609, opened 2016-05-10
Problem: [travis] deploying release artifacts is a manual process

sappo:
Solution: Use travis to deploy these artifacts to github automatically.

The deployment is triggered by tagging on master branch of the
repository. Travis will trigger deployment on the default build
and will in process disable drafts.

For now the results of `make distcheck` are deployed
as well as their md5 and sha1 hash sums.

### DIFF
--- a/zproject_travis.gsl
+++ b/zproject_travis.gsl
@@ -44,6 +44,20 @@ before_install:
 
 # Hand off to generated script for each BUILD_TYPE
 script: ./ci_build.sh
+before_deploy: . ./ci_deploy.sh
+deploy:
+  provider: releases
+  api_key:
+    # To encrypt your access token run: `travis encrypt -r user/repo`
+    secure: <encrypted github access token>
+  file_glob: true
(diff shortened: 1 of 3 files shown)


zproject_travis.gsl
```diff
@@ -44,6 +44,20 @@ before_install:
 
 # Hand off to generated script for each BUILD_TYPE
 script: ./ci_build.sh
+before_deploy: . ./ci_deploy.sh
+deploy:
+  provider: releases
+  api_key:
+    # To encrypt your access token run: `travis encrypt -r user/repo`
+    secure: <encrypted github access token>
```
bluca: Since this has to be manually filled in, won't it be overwritten every time the project is regenerated?

sappo: No, there is a guard on line 16 which prevents zproject from overriding .travis.yml

bluca: Awesome, thanks!

PR #626, opened 2016-06-01
Add compilation with mingw64 using autotools

diorcety:
Enable the compilation on msys2 mingw64 using autotools.

### DIFF
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -250,7 +250,7 @@ AC_MSG_CHECKING([whether to install manpages])
 AC_MSG_RESULT([$$(project.name:c)_install_man])
 
 # Set some default features required by $(project.libname) code.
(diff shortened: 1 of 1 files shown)


zproject_autotools.gsl
```diff
@@ -250,7 +250,7 @@ AC_MSG_CHECKING([whether to install manpages])
 AC_MSG_RESULT([$$(project.name:c)_install_man])
 
 # Set some default features required by $(project.libname) code.
-CPPFLAGS="-D_REENTRANT -D_THREAD_SAFE $CPPFLAGS"
+CPPFLAGS="-D$(PROJECT.PREFIX)_INTERNAL_BUILD -D_REENTRANT -D_THREAD_SAFE $CPPFLAGS"
```
bluca: Shouldn't this be added in the mingw64 section, down below?

diorcety: This is not specific at the end to mingw64. It is indicate that we are building the project (which may not be always the cases for headers)

PR #632, opened 2016-06-06
cd to REPO_DIR before getting dependent repositories

sphaero:


sphaero:
@gotcha you might want to have a look at the last few PRs as I think we don't need docker anymore.

bluca:
Please use the C4 commit message format :-)

Problem: dependency build borken
Solution: cd to REPO_DIR ...

PR #653, opened 2016-07-11
cmake enhancements

sappo:
* Problem: [cmake] release build type uses NDEBUG flag
* Problem: [cmake] summary information is sparse


sappo:
nice a @hintjens merge :)

PR #658, opened 2016-07-13
Problem: [debian] packaging missing dsc and format required by OBS

sappo:
Solution: Add project dsc and format files

### DIFF
--- a/zproject_debian.gsl
+++ b/zproject_debian.gsl
@@ -110,6 +110,8 @@ License: $(project.name)_license
 .   for project.license
  $(string.trim (license.):block                                         )
 .   endfor
+.output "packaging/debian/format"
+3.0 (quilt)
 .output "packaging/debian/rules"
 #!/usr/bin/make -f
 # -*- makefile -*-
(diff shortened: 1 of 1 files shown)


bluca:
Mmh I actually removed this a while ago. It will break the package build, among other things on OBS when using a git service, since dpkg-scansources will find a duplicate dsc.

The dsc is generated at build time by dpkg-buildpackage/debuild

sappo:
Okay, makes sense but OBS didn't start ubuntu/debian build without it.

bluca:
Yes that depends on the type of build used - if you upload manually the tarballs and the artefacts, it will need a dsc.
But instead we can use a source service, for example:

https://github.com/jblunck/obs-service-git_buildpackage
https://github.com/01org/obs-service-git-buildpackage

(the latter supports RPMs too)

We do this at work, and the advantage is that the CI is much simpler. The service will take care of checking out the git tree and generating the artefacts with git-buildpackage -S, so the only thing you have to do from the hook is to kick a remote service run.

I am a bit swamped at work this week, but I can help on IRC, and have a go at it next week if you want

sappo:
Cool, I'll check it out!

bluca:
Great, ping me if you need some pointers.
Also, sorry, the debian/format change was good, and using 3.0 (quilt) is indeed correct :-)

PR #662, opened 2016-07-13
problem: lack of vagrant support

wesyoung:
- support for using vagrant to rapidly spin up a virtualbox/ubuntu environment with all the deps required to use zproject (gsl, etc, similar to dockerfile).

- initial support for using zproject to auto-generate Vagrantfiles using project.xml for other zprojects 

(just getting my feet wet with gsl and zproject, but it appears to "do the right thing", someone may wanna double check the logic though...)

reference:

https://github.com/zeromq/zyre/pull/472
https://github.com/zeromq/zyre/pull/473


c-rack:
Cool!

zproject_vagrant.gsl
```diff
@@ -0,0 +1,77 @@
...
+#   file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+register_target ("vagrant", "packaging for Vagrant")
+
+.macro generate_vagrant_file ()
+.output "Vagrantfile"
+#e -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# This will setup a clean Ubuntu1404 LTS env with a python virtualenv called "pyre" for testing
```
bluca: might want to use the project name rather than hard-coding "pyre"

wesyoung: good call, left over comments from working within pyre.. removed in #663

sappo:
Very neat!

PR #663, opened 2016-07-13
problem: in-accurate comments left in vagrantfile code

wesyoung:


bluca:
Thanks! It also looks like it's a bit borken now:

https://travis-ci.org/zeromq/zproject/jobs/144559305

Could you please have a look at the error that was introduced?

sappo:
@bluca looks like @wesyoung did forget to re-generate zproject with zproject :D

wesyoung:
i'm assuming that's what src/Makemodule.am is for? git was complaining about it so i removed it (and didn't dig too far into .gitignore)... i'll push another PR and see if travis likes it or not (don't merge it).

PR #666, opened 2016-07-15
Problem: search for 'va_list siblings" finds methods that don't meet requirements

eburkitt:
Solution: improve search criteria to require that method with matching name also have a 'va_list' argument

Replaced code erroneously determined zdir_patch.vpath to be a va_list sibling of zdir_patch.path


bluca:
Thanks! For your next contributions, could you please avoid recursive merge nodes and instead rebase your branches on master rather than merging? This makes the history messy and harder to bisect, and cannot be fixed in any way once it's merged.

eburkitt:
Yes, sorry for the mess. Learning Git has been a humbling experience. To make things worse I did some mucking around to make two pull requests before I realized I could select the branches in the UI.

PR #673, opened 2016-07-26
Problem: [obs] zproject has no service file to generate packages on OBS

sappo:
Solution: Generate a service file that pulls sources from github and
extracts packaging files.

Warning: This requires OBS bleeding edge repositories from
openSUSE:Tools to work properly!


zproject_obs.gsl
```diff
@@ -0,0 +1,92 @@
...
+
+.macro target_obs
+.if defined (project.repository)
+.directory.create ('packaging/obs')
+.output "packaging/obs/_service"
+<services>
+  <service name="tar_scm">
+    <param name="url">$(project.repository)</param>
+    <param name="scm">git</param>
+    <!--<param name="version">git</param>-->
```
bluca: Shouldn't this line be removed since it's not used?

zproject_obs.gsl
```diff
@@ -0,0 +1,92 @@
...
+    <param name="outfilename">debian.format</param>
+  </service>
+  <service name="extract_file">
+    <param name="archive">*.tar</param>
+    <param name="files">*/packaging/debian/$(string.replace (project.libname, "_|-"))$(project->version.major).install</param>
+    <param name="outfilename">debian.$(string.replace (project.libname, "_|-"))$(project->version.major).install</param>
+  </service>
+  <service name="extract_file">
+    <param name="archive">*.tar</param>
+    <param name="files">*/packaging/debian/$(string.replace (project.libname, "_|-"))-dev.install</param>
```
bluca: What happens if there's no -dev package, for example when there are no libraries but it's only an executable?
Also I think there's a specific version field for the dev packages, "debian_name", should probably check if that's defined first

bluca:
Really great stuff :-)

zproject_obs.gsl
```diff
@@ -0,0 +1,92 @@
+#   Generate Open Build Servcie recipes for project
```
bluca: "Servcie" -> "Service" :-)

zproject_obs.gsl
sappo: How do we distinct between git and release packages? Maybe we need two service files!?

zproject_obs.gsl
sappo: If there's no -dev package I imagine it will fail miserably :D 

I'll check zproject_debian for the guards!

zproject_obs.gsl
bluca: Good point. I wonder if it's possible to work out whether the last commit is tagged or not, and if not, append something to the revision. There are options to manipulate the version format, but I'm not sure this can be done.
If not I think "git" should be appended, rather than used as a version, otherwise it will cause problems on distributions where the library is already available in the repositories, like Debian and Ubuntu.

The convention when building snapshots for the Debian version is something like:

`<next-tagged-version>~git<data-of-last-commit>-<debian-version`
eg:
`3.0.3~git20160726-1`

This way when the new version is released and uploaded in the distribution it has a higher version (~ is lower than any digit/character when comparing versions)

sappo: The tar_scm service creates this file [_service:tar_scm:_servicedata](https://build.opensuse.org/package/view_file/home:zeromq:czmq/czmq-git/_service:tar_scm:_servicedata?expand=1) which contains the latest commit hash. It think it should be possible to read this file during packaging.

The version convention seem fine to me.

bluca: So it seems it cannot take the next version, but it can take the previous one, and then we can append `+git<DATE>`, which resolves to greater than current versions:

```
dpkg --compare-versions 3.0.2-1 lt 3.0.2+git~20160726; echo $?
0
```

With this versionformat parameter to tar_scm:

`<param name="versionformat">@PARENT_TAG@+git%cd</param>`

The problem is that it creates an invalid version since it includes the "v" from the tag, eg "v3.0.2". I've opened a PR to add a prefix strip option:

https://github.com/openSUSE/obs-service-tar_scm/pull/114

Once that is done and available I can send a PR to do the necessary changes if this sounds sensible.

sappo: Nice!

This solves the git issue. Though how do we distinct between git and release version? Can we compare the parent tag against the version in the dsc file and if they match build a release instead?

bluca: In the release projects we could avoid the `+git%cd` part, so that the version is just the tag.

Although I wonder whether we should have a release project at all. Certainly for distributions that already ship the libraries, like Debian and Ubuntu, I think we shouldn't, as we would be stomping on the maintainer's feet (and some of those feet are mine :-) ), since the versions would be identical, and it would be bound to cause problems for users too due to this.

Might make sense only for distributions where the library is not packaged.

sappo: Yes, that would be bad! It might make sense for projects like zyre and malamute that don't have official packages to have a release package on obs.

bluca: Indeed. As a rule of thumb, I'd say if the package exist for the distribution then don't built releases for it, otherwise it's fine.

PR #674, opened 2016-07-28
Various obs fixes

sappo:


bluca:
Unfortunately it looks like my PR to openSUSE's service will not be merged immediately to fix the @PARENT_TAG@ issue:

https://github.com/openSUSE/obs-service-tar_scm/pull/114

When they do then we can deploy it in our project. Until then it's going to be broken since it uses the full tag with the "v" which breaks it.

sappo:
Okay, then for now obs shows maintainers if their packages are building :)

If I do get the time I'd like to make travis build release packages on obs when tagging a new release. Is the release version correct or do they have the "v" as well?

bluca:
For the release-only builds it should be supported straight from the Github hook, no need to use Travis. If you add a new OBS hook, you can pick the REFS, and with this:

`refs/tags/v*`

It will only trigger on new release tags. it can then point to the release-only OBS project/package.

Unfortunately on the OBS side any version param that uses @PARENT_TAG@ will have the same problem until either my PR or something similar is done.

sappo:
You are right the refs attribute will work. Neat!

Does the _service file need any version or versionformat attribute or is the default sufficient?

bluca:
I think the default if nothing is specified is to append the unix timestamp of the build and nothing else, so no version information, unfortunately.

PR #677, opened 2016-08-10
Problem: zproject cannot cross compile for the Raspberry Pi

sappo:
Solution: Add a cross compile environments which download the cross
compiler, sysroot and dependencies and cross compiles everything.

Tested with Raspbian but should work with Noobs as well!


zproject_rpi.gsl
```diff
@@ -0,0 +1,167 @@
...
+    git clone --depth 1 https://github.com/raspberrypi/tools
+else
+    ( cd tools && git pull --rebase --quiet ) || exit 1
+fi
+
+# Cross build for the Raspberry Pi
+mkdir -p $PWD/tmp
+BUILD_PREFIX=$PWD/tmp
+TOOLCHAIN_HOST="arm-linux-gnueabihf"
+TOOLCHAIN_PATH="/home/ksapper/Workspace/rpi-tools/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/bin"
```
bluca: $HOME ?

sappo: :-D ups! I'll fix it.

Issue #705, opened 2016-10-03
problem: python bindings do not check for relative paths


wesyoung:
i'm working on developing an "embedded" shared object approach to the pthon czmq bindings [still testing] similar to pyzmq (embedded build, requires no system wide libs, the .so files go into the python wheel, etc..).

the issue i've run into is that zproject generates bindings for the MAJOR/MINOR pathing of the shared object (which makes sense), but when we're generating the shared object using python's Extension functionality, there isn't a clear way to ensure we have the proper naming (ie: it comes out as libczmq.so, no .MAJOR), at-least without some more work. obviously we could probably work around that issue, but also seeing an issue where; if you want to interact with the .so, you need to modify the LD_LIBRARY_PATH at runtime, meaning, detecting and re-loading python at times..

so i'm testing the idea of brute force checking two extra paths when loading _project_ctypes.py, assuming a "libczmq.so" may be in either something like "." or "pyczmq". i'm not a huge fan of too many embedded try/excepts, but i'm also not seeing a real clear way to check cdll.* and see if things have already been loaded, or changing the search paths (at-least on UNIX, haven't touched winders yet) without having some magic that manip's the LD_LIBRARY_PATH and reloads the python interpreter.. which, with an embedded module doesn't make a whole lot of sense (esp when you start thinking about things like pyinstaller too...).

i'm open to suggestion and still testing to see if this works in the context of czmq and zyre, but wanted to see if there were other obvious things i'm missing... the patch i'm working on is here:

https://github.com/zeromq/zproject/compare/master...wesyoung:master

i almost want to turn that entire try/except into a function, where someone can set a global var(?) or sources to see if the library was pre-loaded before entering _projectname_ctypes, and called only if the lib isn't loaded (haven't found a great way of checking that either, other than testing for something obvious in one of the data-structures...). give people the ability to hijack the load process if they're trying to do something custom outside of the normal LD_LIBRARY_PATH process, and fall back to this if it's "not already loaded".

wesyoung:
for posterity:

the other thing i've learned, is that you can drop the interpreter into the correct location via os.path.chdir so it searches the -rpath correctly and loads the dependent .so's for this to work (think: libzyre.so requires libczmq.so which requires libzmq.so to function..).

http://stackoverflow.com/questions/4755303/python-cdll-cant-find-module

(google: 'ctypes cdll os.chdir')

[...]

PR #706, opened 2016-10-06
problem: python bindings do not check for bundled shared objects first

wesyoung:
this was the only semi-clean way i was able to solve #705, especially in cases where you may have multiple shared objects embedded within a python library (think: libzyre.so depends on libczmq.so which also depends on libzmq.so and assume they are all under site-packages/czmq and site-packages/zyre instead of LD_LIBRARY_PATH).

i tried testing it both ways in some fresh vm's (loading the original way, and the new way), both seem to check out. once (/if?) this is merged, will submit PR's for new bindings in zyre and czmq.

i'm open to other ways of solving this problem, but most of them deal with modifying LD_LIBRARY_PATH and actually reloading the interpreter, which seems like a bit more complicated code than just dropping in and out the path to see if things load properly..

### DIFF
--- a/zproject_python.gsl
+++ b/zproject_python.gsl
@@ -292,7 +292,10 @@ function generate_binding
     ># $(project.name:c)
     >lib = None
     >try:
-    >    t = os.getcwd()  # get the current dir, temp store it
+    >    # check to see if the shared object was embedded locally, attempt to load it
+    >    # if not, try to load it using the default system paths...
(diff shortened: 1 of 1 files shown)


bluca:
Thanks. Next time if you have additional commits that fix the initial one, please rebase and squash them.

wesyoung:
ya, brain-fart. sorry bout that...

PR #722, opened 2016-12-06
Problem: generated Travis tests do not work out of the box as well as they could, and .gitignore needs updates

jimklimov:
Solution: improve several points so a freshly generated project passes the GitHub/TravisCI tests; rename downloaded Git-sourced projects into `*.git` directories and add those to `.gitignore`

Note: not done for some platform scripts I'm less sure about (android, java, ...) - these might rely on particular directory names and a new suffix might break them. Someone expert in those areas may follow up later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ install_manifest.txt
 # Repositories downloaded by CI integration scripts
 *.git/
 
+# Travis build area
+tmp/
+
 # vagrant
 .vagrant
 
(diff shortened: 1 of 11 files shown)


bluca:
Thanks! Next time, please change the .gsl templates in a separate commit from the local regeneration, so that it's much easier to see what's going on in the history.

@sappo looks after the Android builds so he might be able to help with the directory name changes

PR #724, opened 2016-12-06
Problem: cztx class in czmq is deprecated and removed

malanka:
Solution: change check zctx_test -> zhashx_test

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,7 +431,7 @@ zproject's `use` element defines project dependencies. Model is described in `zp
 
     <use project = "czmq" libname = "libczmq"
         repository = "https://github.com/zeromq/czmq"
-        test = "zctx_test">
+        test = "zhashx_test">
         <use project = "libzmq" />
     </use>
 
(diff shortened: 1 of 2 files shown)


malanka:
And actually projects generated with zproject cannot be compiled, if all necessary CZMQ stuff is manually installed from sources

jimklimov:
@malanka : If you install from sources, do you override path to PKGCONFIG directories to match your local installation? E.g. if you put `libzmq` ecosystem files under `/usr/local/lib`, you should configure your own project with `PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/usr/lib/pkgconfig:${PKG_CONFIG_PATH} ./configure`

(or just `export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/usr/lib/pkgconfig` in your shell before playing)

This helped me when similar behavior bit me.

malanka:
No, I didn't do this. But still, check for a class that does not exist anymore, is bad :)

bluca:
Thanks, missed this when I retired those classes

PR #726, opened 2016-12-07
Problem: builds/cmake/ci_build.sh is delivered not-executable

jimklimov:
Solution: generator should mark it executable


bluca:
For the doc stuff had problems in the past with make distcheck when docs are enabled or disabled, is that all good?

jimklimov:
Tested with the following script:

````
#!/bin/sh

make -k clean
make -k distclean

DISTCHECK_CONFIGURE_FLAGS="PATH=/usr/local/bin:/usr/local/sbin:$PATH PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/usr/lib/pkgconfig --without-docs"

./configure --prefix=/usr/local $DISTCHECK_CONFIGURE_FLAGS DISTCHECK_CONFIGURE_FLAGS="$DISTCHECK_CONFIGURE_FLAGS"
make DISTCHECK_CONFIGURE_FLAGS="$DISTCHECK_CONFIGURE_FLAGS" distcheck
````

So when the base project and distcheck children are configured with the same `--with(out)-docs` settings, distcheck ends good.

bluca:
Nice, thanks!

jimklimov:
I tried a larger matrix (of different settings for parent and dist-child builds), and there is an issue in this case:

````
#!/bin/sh

make -k clean
make -k distclean

DISTCHECK_CONFIGURE_FLAGS="PATH=/usr/local/bin:/usr/local/sbin:$PATH PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/usr/lib/pkgconfig"

./configure --prefix=/usr/local $DISTCHECK_CONFIGURE_FLAGS --without-docs
make DISTCHECK_CONFIGURE_FLAGS="$DISTCHECK_CONFIGURE_FLAGS --with-docs" distcheck

make  dist-gzip dist-zip am__post_remove_distdir='@:'
make[1]: Entering directory '/home/jim/shared/train-0mq/joex'
if test -d "joex-0.1.0"; then find "joex-0.1.0" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "joex-0.1.0" || { sleep 5 && rm -rf "joex-0.1.0"; }; else :; fi
test -d "joex-0.1.0" || mkdir "joex-0.1.0"
 (cd doc && make  top_distdir=../joex-0.1.0 distdir=../joex-0.1.0/doc \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[2]: Entering directory '/home/jim/shared/train-0mq/joex/doc'
make[2]: *** No rule to make target 'joexd.txt', needed by 'distdir'.  Stop.
make[2]: Leaving directory '/home/jim/shared/train-0mq/joex/doc'
make[1]: *** [Makefile:1395: distdir] Error 1
make[1]: Leaving directory '/home/jim/shared/train-0mq/joex'
make: *** [Makefile:1490: dist] Error 2
````

So if the parent was build without docs, and dist(check) wants to distribute them, there is an issue - "dist" enters the original source directory to fetch desired `*.txt` files, but finds none and no recipe either...

Perhaps the `mkman` loop can be moved out of the `if` clauses so the TXTs are always generated.
Alternately, we can avoid EXTRA_DISTing the parts of MAN_TXT that can be generated from sources - both these txts and the manpages are recipe'd in BUILD_DOC=true case.

In any case, the makefile structure "as is" can still fail if INSTALL_MAN=true and BUILD_DOC=false in a freshly cloned or un-tarballed workspace, since there would be no files and no recipes for the manpages. Also a `distcheck` can fail if the build host lacks asciidoc or other tools needed for this conversion :)

bluca:
One thing that should change is that right now, for the Travis-driven release deployment, the docs are not in the make dist. I think this needs to be fixed so that they are in the distributable tarballs uploaded to Github.

Also it would get the default build checking that the docs are built correctly, which right now we don't test.

I'll see how that works and if it crosses in the uses cases you pointed out here.

Thanks for checking!

jimklimov:
Yes, I'm on it now - at least to do pass the distcheck in all combinations.

bluca:
I've got a PR almost up :-)

bluca:
https://github.com/zeromq/zproject/pull/730

jimklimov:
The distcheck part should be fixed by #731

PR #727, opened 2016-12-07
Problem: Cannot add to the documentation some custom defined behaviour

malanka:
Solution: add @interface section to .c file and make mkman to correctly process it


malanka:
The best solution is actually to add a possibility to add some customer things to header file, but it seems difficult. The provided solution is easy to implement :)

PR #736, opened 2016-12-09
Problem: known_projects listed some github upstreams without the .git extension

jimklimov:
Solution: revise the list to always include the extension (less surprise in regenerated ci_build.sh scripts then); also fix imagix => imatix for GSL

### DIFF
--- a/zproject_known_projects.xml
+++ b/zproject_known_projects.xml
@@ -32,23 +32,23 @@
     -->
 
     <use project = "libzmq" prefix = "zmq" debian_name = "libzmq3-dev" redhat_name = "zeromq-devel"
-        repository = "https://github.com/zeromq/libzmq"
+        repository = "https://github.com/zeromq/libzmq.git"
         test = "zmq_init" />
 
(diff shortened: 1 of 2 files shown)


sappo:
@jimklimov we need to double check that the field hasn't been falsely used as URL.

jimklimov:
This should not be a problem - I can open e.g. https://github.com/zeromq/zproject.git in the browser and get redirected to the https://github.com/zeromq/zproject root.

sappo:
Except some smart guy appends the ".git" xD (i.e. Docker and Vagrant)

jimklimov:
You mean lines like these?
````
RUN git clone --quiet $(use.repository).git $(use.project).git
RUN git clone --quiet git://github.com/zeromq/$(project.name:c).git $(project.name:c).git
````

I known the smart guy :) But AFAIK the names should be stripped of the `.git` extension (another bug to tackle, thanks). And some known repos had the extension and some did not.

PS: The benefit of real peer review rather than fast-merge ;)

sappo:
Yes #737

PR #737, opened 2016-12-09
Problem: Appending .git fails if repository already has suffix

sappo:
Solution: Except the .git suffix to be set by the repository
variable

### DIFF
--- a/zproject_docker.gsl
+++ b/zproject_docker.gsl
@@ -40,7 +40,7 @@ RUN sudo ldconfig
 .endfor
 .for use where !optional & !defined (use.tarball)
 WORKDIR /home/zmq
-RUN git clone --quiet $(use.repository).git $(use.project).git
+RUN git clone --quiet $(use.repository) $(use.project).git
 WORKDIR /home/zmq/$(use.project).git
 RUN ./autogen.sh 2> /dev/null
 RUN ./configure --quiet --without-docs
--- a/zproject_vagrant.gsl
+++ b/zproject_vagrant.gsl
@@ -43,7 +43,7 @@ ldconfig
 .endfor
 .for use where !optional & !defined (use.tarball)
 cd /home/vagrant
-git clone --quiet $(use.repository).git $(use.project).git
+git clone --quiet $(use.repository) $(use.project).git
 cd /home/vagrant/$(use.project).git
 \./autogen.sh
 \./configure


jimklimov:
Also makes sense for e.g. locally customized builds where the clone origin may be a directory and so can have or lack an extension. Just use what the XML provides. Sorry for the messup :)

:+1:

sappo:
@jimklimov no worries, this is our way of working :) Keep up the good work :+1:

jimklimov:
I'll try. Can't promise I won't keep up the occasional bad work too, though
;)

2016-12-09 10:58 GMT+01:00 Kevin Sapper <notifications@github.com>:

bluca:
@sappo @jimklimov note that the python binding generation is also broken now:

```
--- a/bindings/python/malamute/_malamute_ctypes.py
+++ b/bindings/python/malamute/_malamute_ctypes.py
@@ -7,7 +7,7 @@ from __future__ import print_function
 import os, sys
 from ctypes import *
 from ctypes.util import find_library
-import czmq
+import czmq.git
 
 # malamute
 lib = None
@@ -57,15 +57,15 @@ lib.mlm_proto_new.argtypes = []
 lib.mlm_proto_destroy.restype = None
 lib.mlm_proto_destroy.argtypes = [POINTER(mlm_proto_p)]
 lib.mlm_proto_recv.restype = c_int
-lib.mlm_proto_recv.argtypes = [mlm_proto_p, czmq.zsock_p]
+lib.mlm_proto_recv.argtypes = [mlm_proto_p, czmq_git.zsock_p]
```

jimklimov:
Not sure what I can do here, my Python is not that strong... Maybe @vyskocilm can chime in?..

Are the "minused" old lines the correct ones? Or on the contrary, should this import match the directory name (now with extension, for automated git clones of prerequisites) somehow?

And this diff probably comes from a selftest or project regeneration (of czmq?), and I see no PR about this... And there are no "git" hits in zproject_python*.gsl so the issue is not something I can easily and stupidly monkey-wrench, sorry :\

bluca:
Yes it's the output of a regeneration from zproject. The Python class should not have a git extension, as it doesn't make much sense, but not sure where is picking that up - from the class template? Or the general one maybe?

bluca:
Urgh can't reproduce it anymore today, not sure what's going on! I'll post back if I can

sappo:
I'll have a look at it tomorrow

Am 11.12.2016 13:51 schrieb "Luca Boccassi" <notifications@github.com>:

sappo:
@bluca I just checked the python bindings. The only way this diff can occur is if the libname and project name is 'czmq.git' or 'czmq_name' respectively.

bluca:
yeah I can't reproduce anymore, not sure how it happened!

PR #738, opened 2016-12-09
Problem: build steps in Travis tests take unknown time, profiling needed

jimklimov:
Solution: to find bottlenecks and prove improvements in Travis integration, the steps (build and test of prerequisites and project itself) should be profiled, at least by adding date'd echo's and time'd invokations

Also fixes another case of imagix=>imatix and some more .git extensions when pulling git clones from internet.

Another minor Problem: CMake travis tests do not report if GitIgnores are clean after build
Solution: Port the block from autotools codepath

### DIFF
--- a/builds/cmake/ci_build.sh
+++ b/builds/cmake/ci_build.sh
@@ -21,23 +21,32 @@ CMAKE_OPTS+=("-DCMAKE_LIBRARY_PATH:PATH=${BUILD_PREFIX}/lib")
 CMAKE_OPTS+=("-DCMAKE_INCLUDE_PATH:PATH=${BUILD_PREFIX}/include")
 
 # Clone and build dependencies
-git clone --quiet --depth 1 https://github.com/imagix/gsl gsl.git
+echo "`date`: Starting build of dependencies: gsl..."
+time git clone --quiet --depth 1 https://github.com/imatix/gsl.git gsl.git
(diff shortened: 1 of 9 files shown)


sappo:
I rather not merge this PR, if you like to profile travis please activate travis on your private repository.

bluca:
Sorry I tend to agree with @sappo, gathering data is important but this kind of work is better done in a private fork.

Enabling Travis is as easy as logging in https://travis-ci.org/profile/jimklimov and enabling the desired repository, and then push again.

jimklimov:
"No value judgments on correct patches"? huh? ;)
Ok, this reluctancy still makes sense; I'll pick out a bug fix I found along the way and resubmit, then.

Just in case - is your reservation in place for both "echo's with date" and "time cmd" runs? Or just against the latter (as more cluttery indeed)?

bluca:
**correct** patches :-P

It's the clutter, it seems to me that this is a worthwhile exercise, but will only be useful right now to look at how things are. It could be merged and then reverted, but then again Travis can be set up in a private repo so there's no point in having every user get the added stuff in all their projects in the meanwhile.

jimklimov:
Ok. For posterity, the bugfixes which were part of this PR were merged as part of https://github.com/zeromq/zproject/pull/739

PR #739, opened 2016-12-09
Problem: zproject CMake differs functionally from default selftest

jimklimov:
Solution: Revise the self-test scripts


jimklimov:
Note: initial PR wanted to fix cmake selftest for zproject to do czmq regeneration too... but then I found it is not called by travis anyway - but rather it is checked that the script regenerated for zproject itself is same as in git.

PR #752, opened 2016-12-16
Problem: make distcheck failed

vyskocilm:
Solution: use dist_bin_SCRIPTS for binaries
see https://www.gnu.org/software/automake/manual/html_node/Scripts.html

### DIFF
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -892,8 +892,7 @@ endif #ENABLE_$(NAME:c)
 .for bin
 .   if first ()
 # Install data into /usr/local/bin
-$(project.name:c)dir = @bindir@
-$(project.name:c)_SCRIPTS = \\
+dist_bin_SCRIPTS = \\
 .   endif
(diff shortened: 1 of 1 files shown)


zproject_autotools.gsl
```diff
@@ -892,8 +892,7 @@ endif #ENABLE_$(NAME:c)
 .for bin
```
jimklimov: @vyskocilm @bluca @sappo : Sanity check - Do we have guaranteed that *all* such `bin`'s are indeed precompiled and git-committed scripts, and not some files generated during build (e.g. from `.in` templates)? The former should be DIST'ed into eventual tarball, but the latter should not (recipe'd build products ought to be rebuilt - and cleaned, while their sources should be DIST'ed).

Also, what was the benefit of not using the project name (and not an easy possibility of a directory other than `bin`) here? Wouldn't `dist_$(project.name:c)_SCRIPTS` achieve the same result while providing unique variable names?

PR #753, opened 2016-12-16
Problem: ci tests runs too long

vyskocilm:
Solution: make distcheck opt-in

### DIFF
--- a/project.xml
+++ b/project.xml
@@ -18,6 +18,7 @@
         redhat              Packaging for RedHat
         ruby                Ruby binding
         travis              Travis CI scripts
+            <option name="distcheck" value="1" /> will enable run of make distcheck instead of make check
         vs2008              Microsoft Visual Studio 2008
         vs2010              Microsoft Visual Studio 2010
         vs2012              Microsoft Visual Studio 2012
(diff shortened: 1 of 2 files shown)


karolhrdina:
Quick discussion with @vyskocilm - Fix coming regarding failing check.

bluca:
I strongly disagree with this change, and would like to revert it.

If it's optional, de-facto it will not run.

Dist-check is fundamental as the github and travis based releases generated by zproject rely on the dist tarball to work properly, and if it's not checked history teaches us that it will break, and then the people doing releases (ie: me :-) ) will have to scramble and do last-minute workarounds

bluca:
How about, as a compromise, to make it opt out instead? So you can disable it in your project if you want (not recommended! :-) )

bluca:
implemented in https://github.com/zeromq/zproject/pull/756 let me know if that causes you any issue

vyskocilm:
Hi,

Wasn't aware about. We're using git archive most of the time, so there's a
little value and a lot of hurdles for it's.

So your change is good for us.

Dne 16. 12. 2016 16:04 napsal uživatel "Luca Boccassi" <
notifications@github.com>:

PR #757, opened 2016-12-19
Problem: generated ci script does not work for tntdb

vyskocilm:
Solution: add a builddir to zproject use model


sappo:
What is tntdb?

malanka:
https://github.com/maekitalo/tntdb

PR #761, opened 2016-12-22
Problem: projects which choose to not track CMake recipes fail distcheck

jimklimov:
Solution: have ./configure script detect the presence of $srcdir/CMakeLists.txt and based on that - makefile should list them among EXTRA_DIST (or not), effectively making CMake support optional (now hardcoded among default targets for project.xml)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -22,10 +22,15 @@ noinst_PROGRAMS =
 check_PROGRAMS =
 noinst_LTLIBRARIES =
 TESTS =
+EXTRA_DIST =
 
-EXTRA_DIST = \
-    CMakeLists.txt \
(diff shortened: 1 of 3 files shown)


sappo:
Is the cmake flag enabled by default?

jimklimov:
yes, it is part of zproject.gsl : 

````
    assume_target ("autotools")
    assume_target ("cmake")
````

jimklimov:
Verified with a project where the CMakeFiles was moved away : the autotools distcheck passed https://travis-ci.org/eaton-bob/joex/jobs/186048485 and the cmake job failed for some reason ;) https://travis-ci.org/eaton-bob/joex/jobs/186048487

PR #777, opened 2017-01-02
Problem: When building tarballs, zproject_travis.gsl returns to ".."

jimklimov:
…rather than to `$BASE_PWD` (this can have adverse effect if the project has a defined subdir with sources for build - e.g. not the tarball root)

Solution: Carry over BASE_PWD from git-build clause


sappo:
In other places we use `pushd` and `popd` commands instead. Those might be a bit cleaner.

PR #786, opened 2017-01-02
Problem: unavoidable "configure --quiet" obstructs justice

jimklimov:
Solution: when travis tests go south, more insight can be desired - so make the quiet configure operation default, but togglable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,12 @@ sudo: false
 
 # Set CI_TIME=true to enable build-step profiling in Travis
 # Set CI_TRACE=true to enable shell script tracing in Travis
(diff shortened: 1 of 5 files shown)


bluca:
What about using the TRACE variable instead of a new one? So that there's one less knob, and they would most likely be used together anyway when debugging

jimklimov:
Using one knob for several things is not a minimal solution :)

Teammates generally prefer to have as little output in logs as they can navigate and glance over, so it sounds like a valid argument in favor of two knobs (script tracing is quite noisy and not everyone likes to enable it).

bluca:
Ok, makes sense thinking about it

PR #794, opened 2017-01-04
Problem: zproject tracked sources differ from what zproject currently generates for others

jimklimov:
Solution: fix the typo

### DIFF
--- a/doc/mkman
+++ b/doc/mkman
@@ -104,6 +104,8 @@ sub generate_manpage {
     }
     die "Can't open '$source'"
         unless open (SOURCE, $source);
+    # NOTE: This duplication of lines is needed for proper work:
+    $_ = <SOURCE>;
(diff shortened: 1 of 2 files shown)


doc/mkman
```diff
@@ -97,13 +97,15 @@ sub generate_manpage {
...
-        if ($header == $source) {
+        if ($header eq $source) {
             $header .= "c";
         }
         $source .= "c"; # Retry for a *.cc filename
     }
     die "Can't open '$source'"
         unless open (SOURCE, $source);
+    # NOTE: This duplication of lines is needed for proper work:
+    $_ = <SOURCE>;
```
bluca: I thought this was very weird, then I realized it's Perl

jimklimov: Troll detected! :D

PR #796, opened 2017-01-05
Problem: comment in control file breaks BuildService parser

vyskocilm:
Solution: drop it

### DIFF
--- a/zproject_debian.gsl
+++ b/zproject_debian.gsl
@@ -51,7 +51,6 @@ Build-Depends: debhelper (>= 9),
 .endif
 .endfor
 .if systemd ?= 1
-# necessary for systemd.pc to get unit install path
     systemd,
     dh-systemd,
 .endif


sappo:
@bluca added this comment to remind him/us not to remove systemd in the future. Is there any way to retain the comment without breaking the BuildService?

jimklimov:
@sappo Is there a GSL comment tag?

sappo:
@jimklimov excellent idea!

bluca:
By build-service you mean OBS? It doesn't cause problems for example: https://build.opensuse.org/package/show/home:zeromq:git-stable/czmq

bluca:
GSL comment works too, just curious about the failure since I haven't seen it

jimklimov:
Michal is not around ATM, but I can guess that our (internal) OBS may be different both due to age and local tinkering/forking.

vyskocilm:
Our problém was that with a comment dh-systemd wasn't installed. Once I
removed the line, system started to work again

Dne 5. 1. 2017 15:15 napsal uživatel "Jim Klimov" <notifications@github.com

PR #801, opened 2017-01-05
Problem: zproject allows to constrain only minimal version of dependency

jimklimov:
Solution: add support for optional maximum and next known-incompatible versions

Verified to produce reasonable output in both `configure.ac` and `*-pc.in` files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,6 +428,10 @@ Model is described in `zproject_known_projects.xml` file:
         repository = "https://github.com/zeromq/libzmq.git"
         test = "zmq_init" />
 
+    <!-- Note: if your project requires an older CZMQ (e.g. if you need
+        `release = "v3.0.2"`), you may need to `test = "zctx_test"`.
+        Also note that you can instead require particular package
+        version (as reported by pkg-config records). -->
     <use project = "czmq" libname = "libczmq"
         repository = "https://github.com/zeromq/czmq.git"
         test = "zhashx_test">
--- a/README.md
+++ b/README.md
@@ -250,6 +250,10 @@ zproject's `project.xml` contains an extensive description of the available conf
         The travis ci will use the installed packages when building instead of
         rebuilding if available.
     <use project = "zyre" min_major= "1" min_minor = "1" min_patch = "0" />
+    <use project = "czmq"
+        min_major= "3" min_minor = "0" min_patch = "2"
+        next_incompatible_major = "4"
+        />
     <use project = "uuid" optional= "1" implied = "1" />
     <use project = "myfirstlib" repository = "http://myfirstlib.org/myfirstlib.git" />
     <use project = "mysecondlib" tarball = "http://mysecondlib.org/mysecondlib-1.2.3.tar.gz" />
--- a/project.xml
+++ b/project.xml
@@ -74,6 +74,10 @@
         The travis ci will use the installed packages when building instead of
         rebuilding if available.
     <use project = "zyre" min_major= "1" min_minor = "1" min_patch = "0" />
+    <use project = "czmq"
+        min_major= "3" min_minor = "0" min_patch = "2"
+        next_incompatible_major = "4"
+        />
     <use project = "uuid" optional= "1" implied = "1" />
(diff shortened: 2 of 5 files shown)


bluca:
Same constraint should be set in the rpm and deb packages

jimklimov:
Possibly, sort of. There is a problem of package versions != source versions that happen in some projects.
Also, I only looked for hits of "min_version" and "min_major" precedents ;)

jimklimov:
BTW, is there any good reason to (always) require the pkg-config'ed package (if set so) and avoid tests? Would a toggle make sense (to fall back to testing rather than blunt warn/error)?

jimklimov:
Well... there are things like requiring features from (or close to) Git HEAD which for zeromq were tagged as 4.2.0 for a while - without a release yet named so, IIRC. So there were (locally made) packages versioned like `4.2.0.$timestamp` or some such.

But other than nits like that, I'm not sure zmq ecosystem had such debauchery. There are third-party projects that can be relied on with the `<use>` tag in a `project.xml`, and the internet is full of randomly bad things ;)

bluca:
It checks for pkg-config first as that's the easiest, sanest and most maintainable check, and then it does the test only as a fallback if pkg-config is not there

jimklimov:
But otherwise, feel free to improve :)

Perhaps in a similar vein, default the package versions to pkg-config (source) versions, or override with similar component tags.

Thinking of it, there are also distro conventions that make version strings complicated, like truncating to `MAJ.MIN+respin@arch%whatever` which is what an `apt` or `rpm` might look for but is tangentially related to source version :(

jimklimov:
> It checks for pkg-config first as that's the easiest, sanest and most maintainable check, and then it does the test only as a fallback if pkg-config is not there

No, in fact it does the fallback only if versions are not specified (so just looking for presence of SOME version >= 0.0.0 of the package's pkgconfig metadata).

bluca:
The configure run on zyre:

```
checking for libzmq... yes
checking for czmq... yes
checking whether to build documentation... no
```

If I mangle zyre's configure.ac so that pkg-config returns false:

```
diff --git a/configure.ac b/configure.ac
index 81cfc7f..45f83b4 100644
--- a/configure.ac
+++ b/configure.ac
@@ -127,7 +127,7 @@ fi
 
 was_czmq_check_lib_detected=no
 
-PKG_CHECK_MODULES([czmq], [libczmq >= 0.0.0],
+PKG_CHECK_MODULES([czmq], [libcsszmq >= 0.0.0],
     [
         PKGCFG_LIBS_PRIVATE="$PKGCFG_LIBS_PRIVATE $czmq_LIBS"
     ],
```

Then it falls back to doing the zhashx_test:

```
checking for czmq... no
checking for zhashx_test in -lczmq... yes
checking whether to build documentation... no
```

And finishes correctly, as intended
What problem are you seeing with the checks?

jimklimov:
You have the codepath for `>= 0.0.0` above. If the version value is `0.0.0` there is the fallback. If you looked for `<use project = "czmq" min_major = "1" ...>` (and regenerated the zproject) this fallback check will no longer be there, just the warning or error based on optionality of the dependency.

I am not certain it is always good, e.g. might break on systems without pkg-config (tool, or not installed files), so wondered if a toggle to always have the fallback makes sense.

bluca:
Ah I see what you mean now

I guess the problem is that you can't really do a version check any other way, right? How could it be implemented otherwise?

jimklimov:
Not sure, just thinking aloud... :)

Perhaps with the toggle implemented and set, the project would look for requested pkg-config dependency versions, but if those are not available (and if asked to) - fall back to looking at libraries and headers that contain a required routine. Example might be czmq having `zctx_test` vs. `zhashx_test`, which does allow to differentiate versions effectively (by contracts/APIs, which often is what really matters).

bluca:
So using the test as a sort of version check? That might work. In this case simply removing the ```. if use.min_version``` should be enough. I wouldn't add a new toggle, we already have a lot of options, and always checking pkg config first and then falling back to the test makes sense

PR #802, opened 2017-01-05
Problem: Self-test skeleton is inconsistent

jimklimov:
Investigation step 1 - visibility : in projname_selftest.c(c) report what features a selftested class has (draft/stable, private/public)

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -202,11 +202,15 @@ typedef struct {
 static test_item_t
 all_tests [] = {
 .for class where !draft & selftest & private ?<> 1
+.   if first ()
+// Tests for stable public classes:
+.   endif
     { "$(class.c_name)", $(class.c_name)_test },
 .endfor
 .for class where draft & selftest & private ?<> 1
 .   if first ()
 #ifdef $(PROJECT.PREFIX)_BUILD_DRAFT_API
+// Tests for draft public classes:
 .   endif
     { "$(class.c_name)", $(class.c_name)_test },
 .   if last ()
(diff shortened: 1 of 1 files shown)


jimklimov:
Seems this is linked to https://github.com/zeromq/zproject/commit/e63ccee20f910f0656eda75fc69800b749549b84 and https://github.com/zeromq/zproject/commit/d44194965e5ccd61fa9d9d9a636a62581fa67adb

At least for the help text the behavior is inconsistent (it lists all classes, not only those that can be tested).

jimklimov:
@bluca : comment to those commits says that the public class selftest should invoke private class selftests; is this enforced somehow by zproject?

bluca:
It's not enforced by zproject as unfortunately I couldn't think of a sane way to do it. Suggestions welcome!

The issue is that the self test binary is, well, a binary. So it just links to the shared library. To be able to see a symbol, it has to be public

zproject_class.gsl
```diff
@@ -276,9 +280,24 @@ main (int argc, char **argv)
...
+.   if class.private ?<> 1
+public  \
+.   else
+private \
+.   endif
+API; selftest is \
+.   if !class.selftest
+NOT \
+.   endif
+implemented");
```
bluca: Doesn't the code check that the test is available anyway before adding it? Isn't this output superfluous then?

jimklimov: In this help text - it just goes `for class` with no filters

bluca: Ah I see, that was broken - what about adding a filter rather than the extra output? So the text is less verbose, and correct nonetheless

jimklimov:
Looking at code, I wonder if the same generated skeleton (array and loops) as made for selftest binary can be put into the library and exposed as one method, e.g.  `test_private_classes()` and called from the binary?

bluca:
That should work. It would need a new module, but the existing main header would do for the symbol definition.

The only alternative I can think of is using GSL to find for each private class find the first public class which is using it and call the method from there. Basically automate what I've done manually as a workaround.

zproject_class.gsl
bluca: Done in https://github.com/zeromq/zproject/pull/809

zproject_class.gsl
```diff
@@ -276,9 +280,24 @@ main (int argc, char **argv)
         else
         if (streq (argv [argn], "--list")
         ||  streq (argv [argn], "-l")) {
-            puts ("Available tests:");
+            puts ("Available tests (note that only stable classes are exposed to test):");
```
jimklimov: Ah, sorry. Should have been public :\

PR #803, opened 2017-01-05
Problem: generated debian control file ended a list with comma

jimklimov:
Solution: shuffle around the required package names so the end of list is definitive

### DIFF
--- a/packaging/debian/control
+++ b/packaging/debian/control
@@ -16,8 +16,9 @@ Maintainer:     zproject Developers <zeromq-dev@lists.zeromq.org>
 Standards-Version: 3.9.7
 Build-Depends: debhelper (>= 9),
     pkg-config,
-    dh-autoreconf,
     generator-scripting-language,
+    dh-autoreconf
+
 Build-Depends-Indep: asciidoc,
                      xmlto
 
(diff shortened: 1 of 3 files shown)


bluca:
This is not actually a problem, in fact it's supported. It means less churn (1 line diff) when adding new dependencies.

jimklimov:
Nice to hear, thanks. But anyhow, the result looks cleaner this way =D

bluca:
Well it will not look clean anymore as things are added :-) I wrote it like that intentionally, and it's pretty much common to use that format among Debian devs.

Also, the change as-is is broken anyway - there shouldn't be a space between the Build-Depends and Build-Depends-Indep lines, so the build is broken:

```
$ debuild -uc -us -S
 dpkg-buildpackage -rfakeroot -d -us -uc -S
dpkg-buildpackage: source package czmq
dpkg-buildpackage: source version 4.0.3
dpkg-buildpackage: source distribution UNRELEASED
dpkg-buildpackage: source changed by czmq Developers <zeromq-dev@lists.zeromq.org>
 dpkg-source --before-build czmq-upstream
dpkg-source: error: syntax error in czmq-upstream/debian/control at line 26: block lacks the 'Package' field
dpkg-buildpackage: error: dpkg-source --before-build czmq-upstream gave error exit status 25
debuild: fatal error at line 1376:
dpkg-buildpackage -rfakeroot -d -us -uc -S failed
```

So I'll revert this if you don't mind

jimklimov:
Or we can remove the line, PR pending ;)

jimklimov:
Regarding churn: The additional packages can be prepended in the stack (so this dh-autoreconf is always last, and those have a trailing comma) so no extra churn in diffs.

PR #804, opened 2017-01-05
Revert "Problem: generated debian control file ended a list with comma"

bluca:
Reverts zeromq/zproject#803

Breaks package build, and will cause more churn as deps are added


jimklimov:
Alternative is #805 to try not breaking builds. The additional packages can be prepended in the stack (so this dh-autoreconf is always last, and those have a trailing comma) so no extra churn in diffs.

bluca:
That's fine for now, but at some point I'll wrap-and-sort all the files (they are a bit messy now) and I'll reintroduce that

PR #807, opened 2017-01-05
Problem: different systemd units can be delivered by projects

jimklimov:
Solution: rather than fixed `$name.service` strings, allow packaging of certain globs on top of this.

TODO: Test resulting markup with some packaging system (OBS?); by docs globs should be supported.

### DIFF
--- a/zproject_debian.gsl
+++ b/zproject_debian.gsl
@@ -195,7 +195,7 @@ usr/bin/$(bin.name)
 .# generate service file names
 .   for project.main where main.service ?= 1
 etc/$(project.name)/$(main.name).cfg
-lib/systemd/system/$(main.name).service
+lib/systemd/system/$(main.name){,@*}.{service,timer,path,target,socket}
 .   endfor
 .endif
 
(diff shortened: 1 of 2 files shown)


bluca:
I can give it a spin

jimklimov:
Thanks, and I can go home before I miss another train :)

bluca:
Unfortunately this doesn't work as dh_install fails given not all variants are there:

```
   dh_install
cp: cannot stat ‘debian/tmp/lib/systemd/system/malamute.timer’: No such file or directory
dh_install: cp --reflink=auto -a debian/tmp/lib/systemd/system/malamute.timer debian/malamute//lib/systemd/system/ returned exit code 1
debian/rules:40: recipe for target 'binary' failed
make: *** [binary] Error 2
```

As an alternative, perhaps use GSL to check if each combination exists and add them at generation time

bluca:
Or perhaps something like:

```
lib/systemd/system/$(main.name)*
```

PR #811, opened 2017-01-06
Relax the glob for systemd unit types

jimklimov:
Follow-up to #807

### DIFF
--- a/zproject_debian.gsl
+++ b/zproject_debian.gsl
@@ -195,7 +195,7 @@ usr/bin/$(bin.name)
 .# generate service file names
 .   for project.main where main.service ?= 1
 etc/$(project.name)/$(main.name).cfg
-lib/systemd/system/$(main.name){,@*}.{service,timer,path,target,socket}
+lib/systemd/system/$(main.name){,@*}.{service,*}
 .   endfor
 .endif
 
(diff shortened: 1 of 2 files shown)


bluca:
@jimklimov just regenerated malamute with this change, but the build on OpenSUSE, RHEL and CentOS fails (strangely not on Fedora):

```
[  109s] RPM build errors:
[  109s]     File not found by glob: /home/abuild/rpmbuild/BUILDROOT/malamute-1485092044.246fc77-41.1.x86_64/usr/lib/systemd/system/malamute{,@*}.{service,*}
```

https://build.opensuse.org/package/show/home:zeromq:git-stable/malamute

Could you please have a look? I'm not a huge RPM expert

jimklimov:
Strange indeed... perhaps there are different versions or settings of RPM with varying capabilities.

From those build logs, it does `install` the service file into where it looks by the glob later, e.g.

````
install : /home/abuild/rpmbuild/BUILDROOT/malamute-1485092044.246fc77-41.1.x86_64/usr/lib/systemd/system/malamute.service
errglob : /home/abuild/rpmbuild/BUILDROOT/malamute-1485092044.246fc77-41.1.x86_64/usr/lib/systemd/system/malamute{,@*}.{service,*}
````

I think the short-term solution would be to drop the complex glob after all (debian doesn't like it either) and use directly names like `file.service` and perhaps `file@*.service` (if appropriate template or file is detected in workspace during zproject run, and/or add an option to the project to specify if it has a single- or multi-instance service... or perhaps both for different purposes)... And then it gets hairy but "right" to add similar supports for other unit types - in project.xml or based on presence during the zproject run, whenever it gets needed ;)

PR #812, opened 2017-01-06
Problem: wrong name of install file

vyskocilm:
Solution: use the same number as project version


bluca:
The binary package of a library has to be named after the SONAME version, not the project version, as per Debian policy:

https://www.debian.org/doc/debian-policy/ch-sharedlibs.html

vyskocilm:
it must be consistent, so install file and package name in control file
 must be the same. We can put 0 everywhere if no abi is defined

Dne 6. 1. 2017 11:46 AM napsal uživatel "Luca Boccassi" <
notifications@github.com>:

bluca:
agh I missed one then, let me fix that

Issue #806, opened 2017-01-05
Problem: doc/Makefile is wrong for complex projects


jimklimov:
Lest I forget before the night finally dawns on me and dims my mind, while regenerating a project I hit a few issues with newly generated `doc/Makefile.am`:

* mixup of minuses and underscores in filenames : de-facto, MAN1 pages (for public `<main>` items => programs) are listed with minuses (`$(name)`) and MAN3 ones are with underscores (for public `<class>` items, with `$(name:c)`); recipes (txt from C/CC) generate `$(builddir)/...` filename with minuses (and even spaces in some cases) - so it breaks for MAN3 targets
* there are checks in GSL on whether a class/main name is same as a project overall, and based on that MAN7 (documentation for `$(project.name)`) is cleared. IMHO this is wrong - one text can be programming doc (generated), and another a usage or rationale doc (manually maintained) - but should land into different TXT filenames (e.g. `filename.man7.txt`?)... skeletons for the two cases are sufficiently different too.
* for a project with several `<main>` tags, several entries were correctly listed in the MAN1 list, but only one skeleton seems to have been generated by a zproject GSL run (and even that - for the man7 codepath according to contents, but conflicting with the filename gitignored for man1 and so not easily `git add`able) 

So this spaghetti must be untangled :)

jimklimov:
In the end, issue solved with PR #814 + #815 ; note that the project-overview (man7) document is now always provided (even if name collides with class or main) with a different extension (`adoc` for the git-trackable file, and `xml7` for intermediate data).

bluca:
Good stuff :+1:

bluca:
@jimklimov there's a problem with CZMQ:

```
make[1]: Entering directory `/home/travis/build/bluca/czmq/czmq-4.0.3/_build'
Making all in doc
make[2]: Entering directory `/home/travis/build/bluca/czmq/czmq-4.0.3/_build/doc'
make[2]: *** No rule to make target `zactor.doc', needed by `doc'.  Stop.
```

https://travis-ci.org/bluca/czmq/jobs/189805961

Could you please have a look when you have time?

jimklimov:
Ok, though probably not before next week starts.

bluca:
Sure no problem, thank you

jimklimov:
Currently known problems are all addressed, closing the issue (again ;) ).

PR #821, opened 2017-01-09
Problem: dummy recipes for *.doc generation were not processed

jimklimov:
Solution: seems a ".txt.doc" recipe must have some payload to be processed, so make it "@true" - now "make filename.doc" does not complain about "No rule to make..." but does the proper job.

This should fix the remainder of issue #806 (defect found while testing an earlier solution, see comments there).


bluca:
Thanks, I'll test this in CZMQ in a short while

jimklimov:
At the very least, this issue broke one of our components, and this fix fixed it - both on command-line and in Travis :)

bluca:
Looks good on CZMQ as well: https://travis-ci.org/bluca/czmq/builds/190198392 thanks!

PR #823, opened 2017-01-09
Problem: got issues building against lua

jimklimov:
Solution:
* enable fallback testing if pkg-config data not present
* enable opting-out of redhat and debian package install/req/check by providing empty string as the name

Usecase that passes build is:

````
    <!-- Note: the optional flag causes "configure" script to not fail
         if pkg-config metadata (or tests) do not succeed on a distro;
         the project build would fail if no LUA-5.1 library if present.
         This also causes some duplicates in generated code, but these
         "should" be harmless in practice. Also, these packages are not
         generated into the list of "required" for a build in spec/dsc
         files and have to be plugged there manually. -->
    <use project = "lua-5.1-redhat" libname = "lua" prefix="lua"
        optional = "1" am_lib_macro = "LUA_5_1"
        min_major = "5" min_minor = "1" mit_patch = "0"
        debian_name="" redhat_name="lua-devel"
        test = "lua_close"
    />
    <use project = "lua-5.1-debian" libname = "lua5.1" prefix="lua"
        optional = "1" am_lib_macro = "LUA_5_1"
        min_major = "5" min_minor = "1" mit_patch = "0"
        debian_name="liblua5.1-0-dev" redhat_name=""
        test = "lua_close"
    />
````

### DIFF
--- a/zproject_cmake.gsl
+++ b/zproject_cmake.gsl
@@ -540,14 +540,18 @@ CMAKE_OPTS+=("-DCMAKE_INCLUDE_PATH:PATH=${BUILD_PREFIX}/include")
 # Clone and build dependencies
 [ -z "$CI_TIME" ] || echo "`date`: Starting build of dependencies (if any)..."
 .for use where defined (use.tarball)
-.   if defined (use.debian_name)
-if ! ((command -v dpkg-query >/dev/null 2>&1 && dpkg-query --list $(use.debian_name) >/dev/null 2>&1) || \\
+if \
+.   if defined (use.debian_name) 
+.       if !(use.debian_name = '')
+\! (command -v dpkg-query >/dev/null 2>&1 && dpkg-query --list $(use.debian_name) >/dev/null 2>&1) || \\
+.       endif
 .   elsif defined (use.libname)
-if ! ((command -v dpkg-query >/dev/null 2>&1 && dpkg-query --list $(use.libname)-dev >/dev/null 2>&1) || \\
+\! (command -v dpkg-query >/dev/null 2>&1 && dpkg-query --list $(use.libname)-dev >/dev/null 2>&1) || \\
 .   else
-if ! ((command -v dpkg-query >/dev/null 2>&1 && dpkg-query --list $(use.project)-dev >/dev/null 2>&1) || \\
+\! (command -v dpkg-query >/dev/null 2>&1 && dpkg-query --list $(use.project)-dev >/dev/null 2>&1) || \\
 .   endif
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by jimklimov** [comment]
Note: a prettier solution might be looping over a list of variants, but it proved much too complex for a day's work.

File: zproject_cmake.gsl [hunk 1]
```diff
@@ -540,14 +540,18 @@ CMAKE_OPTS+=("-DCMAKE_INCLUDE_PATH:PATH=${BUILD_PREFIX}/include")
 # Clone and build dependencies
 [ -z "$CI_TIME" ] || echo "`date`: Starting build of dependencies (if any)..."
 .for use where defined (use.tarball)
-.   if defined (use.debian_name)
-if ! ((command -v dpkg-query >/dev/null 2>&1 && dpkg-query --list $(use.debian_name) >/dev/null 2>&1) || \\
+if \
+.   if defined (use.debian_name) 
+.       if !(use.debian_name = '')
```
bluca: Is this necessary? If the filename is empty the check will simply fail, don't think that's a problem

jimklimov: No, in practice "defined" and "empty" are two different states. In the usecase posted into the PR, above, I want to test for `lua` or `lua51` packages - because OS distributors want to make our lives a hell, and we hardcode a single expected package or lib name during project-source generation (hint - this is the real improvement point... or providing a custom `ac-find-lua.m4` for this non-generic case). In this case one test is essentially for DEB and other is for RPM systems. 

And I don't want the pre-generated and fixed-forever packaging manifests referring to names not present in the current OS nor to produce empty tokens nor to produce duplicate entries - so I found that an explicit empty value allows to skip that. Also note we only have concept of one debian and one redhat to rule them all by design here - so all ubuntus and travis testers and opensuses etc. must bow and comply to the one and only true name we utter in known-platforms or the project.xml,

jimklimov: No, in practice "defined" and "empty" are two different states. In the usecase posted into the PR, above, I want to test for `lua` or `lua51` packages - because OS distributors want to make our lives a hell, and we hardcode a single expected package or lib name during project-source generation (hint - this is the real improvement point... or providing a custom `ac-find-lua.m4` for this non-generic case). In this case one test is essentially for DEB and other is for RPM systems. 

And I don't want the pre-generated and fixed-forever packaging manifests referring to names not present in the current OS nor to produce empty tokens nor to produce duplicate entries - so I found that an explicit empty value allows to skip that. Also note we only have concept of one debian and one redhat to rule them all by design here - so all ubuntus and travis testers and opensuses etc. must bow and comply to the one and only true name we utter in known-platforms or the project.xml.

PR #824, opened 2017-01-09
Problem: underscore is not valid debian package name

vyskocilm:
Solution: call string.replace in travis script, like debian do

cc @jimklimov


jimklimov:
Well, if debian packaging also does follow this pattern (so conversion ends up in a valid package name that is likely to exist in apt repos) - LGTM then :)

Issue #810, opened 2017-01-06
Problem: support wanted for systemd tmpfiles to spawn data directories and other service enhancements


jimklimov:
Solution:
- [ ] add a project tag to enable tmpfiles integration (with a generated .in template => configure/extra-dist)
- [ ] deliver the file into systemd location according to config
- [ ] in service config file (or unit file?) set the workdir to specified location
- [ ] optionally set specified username and group to own the FS objects, and add to service template to run as
- [ ] set which target milestone this service is part of (if not default multi-user)

vyskocilm:
There was an installation model long time ago. I dropped it in #539 as it become hard to understand and messy. @hintjens told me, file installation is **HARD** problem and a lot of people failed to solve it. The more properties you add (I heard about groups/owners) it will become more complex and less platform independent, harder to use and impossible to use it correctly. See automake. foo_bar_SCRIPTS EXTRA_DIST, ham_BIN, .... I even have no idea why there are so many options there.

I am not saying that we should not do it, just to have a good model initially.

cc @bluca @sappo

bluca:
I agree it could get very hairy if it becomes a matrix of various bits and pieces from build systems, packaging, etc

vyskocilm:
@bluca take a look on #825 - maybe it's the beer what is talking instead of me, but I consider my design as briliant :-D I tried to solve this puzzle several times and ... I found the way to experiment easily and safelly w/o adding to much burden to the rest of zproject.

jimklimov:
Not exactly continuing on-topic, but well yes, packaging is "hairy" indeed, and it is something that automated projects should automate at build-time, not generation time. For example, one (more) annoyance with practical use of zproject is that the debian/redhat/travis files are pre-generated and dependency package names are set in stone, for all distros a project might be built on. And different distros, while using similar packaging software, often use very different package names (e.g. more and more I thing that an additional `travis_name` for dependency packages is warranted, even if defaulted to `debian_name` for most packages).

And it is not a problem that zproject, or indeed any application software project (whether generated or manually crafted), should be *solving* alone once and for all with static files, but rather one it can well help with - e.g. with `.in` templates populated by `configure` based on what packages it actually finds on the build system and/or falls-back as hardcoded.

vyskocilm:
BTW: see our discussion about installation model with Pieter: https://github.com/zeromq/zproject/pull/373

PR #825, opened 2017-01-09
Problem: can't install additional files via project.xml

vyskocilm:
Solution: introduce zinstall, the zero model for file installation

The beauty of this model (as opposite of PR#539) is that it impose clear
protocol between model and rest of the generators. This protocol are
files in builds/zinstall/, which are then included in generated files.

This has two main advantages over older attempts
1. installation model fully implemented in ONE place
2. because of the protocol, there can be more models w/o any impact on
backend

Design goals are stated in README.txt, but let's repease
1. KISS - the less options the better
2. No conditionals - please don't create an another half-baked XML
programming langugage like ant
3. Real problems - don't add more options because you can
... or you will get replaced by zinstall2 ;-)

foundation for closing #810

cc @bluca @sappo @jimklimov @karolhrdina 

extra-cc @hintjens  - sorry, but I was so slooow with solving this :-(

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -432,6 +432,28 @@ Using autotools or CMake, you can specify --with-drafts to enable draft APIs, an
 make uninstall
 ```
 
+## Additional files
+Installation of third party files is a *hard* problem. It is not platform
+independent, became hard to maintain and impossible to use correctly. One of
+zproject's goals is a simplicity. There is a simple installation model
+
+### Design goals
+* KISS, less configuration options the better
+* no conditionals in the model, those SHALL be handled in background
+* each option solves a REAL problem, avoid extending it because you can
+
+### Example
+```
+    <main name = "MAIN">
+        <install type = "systemd-tmpfiles" />
+    </main>
+```
+
+This will add install information about systemd tmpfiles.d configuration files
+to autotools, packaging, and so. The resulting file
+/usr/lib/tmpfiles.d/MAIN.conf will be installed only if configure was called
+with --with-systemd-units.
+
 ## Notes for Writing Language Targets
 
 This is the general form of a target:
--- a/zproject.gsl
+++ b/zproject.gsl
@@ -172,6 +172,9 @@ gsl from "zproject_bench.gsl"
 gsl from "zproject_git.gsl"
 gsl from "zproject_valgrind.gsl"
 
+# zinstall must run before all the targets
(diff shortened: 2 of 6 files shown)


vyskocilm:
I must say I am extremly proud of it! I wanted to create something like zinstall project initially. This design goal have helped me to streamline the implementation and to define builds/zinstall protocol. I am sure that pieter will love it. Unlike my initial attempt. Some lessons take their time!

bluca:
Thanks, this is a very interesting idea! Let's see how it works out in the various projects

sappo:
@vyskocilm this looks like magic to me. Can you explain to me what's happening? Does the user have to manually create the files in the zinstall directory?

sappo:
@vyskocilm I've taken a closer look at your code and the name `zinstall` is very deceiving as you're not installing anything but merely providing code snippets to be inserted somewhere else. I'm not convinced yet because I don't understand the Problem. Let's start by what are systemd tempfiles? And why do I need them in zproject?

vyskocilm:
@sappo - I made an example in a meanwhile, see https://github.com/vyskocilm/zinstall/tree/master/builds/zinstall

vyskocilm:
@sappo maybe it'd be better to have design discussion elsewhere, but well

Our problem is that our daemons store their state as a plain files in some runtime directory, which is somewhere in /var/lib/$project/$daemon. So normally there's no way how daemon can create it in runtime. And I belive expecting daemon to create own runtime dir is against all the modern expectations for containers and so. Our solution is to distribute systemd-tmpfiles.d config file, so the runtime directory will be managed by systemd.

Note that zproject_install.gsl is not limited to this type, it can be extended. But as this the real problem we're facing, zproject_install.gsl solve excactly this and nothing more.

Zproject itself has no capability to express installation of additional files. Mainly because it's not an easy problem to solve. See the #539 how ugly was my previous attempt to solve it. It was implemented everywhere in project and as we've have three generations/iterations of approach, those were handled everywhere. This is an attempt to solve this problem again, but without adding too much burden to the rest of zproject scripts. The zinstall_include lines are the protocol between scripts and install models.

BTW: if you don't like builds/zinstall - well, naming is a hard problem, so I'll welcome your pull requests ;-)

BTW2: I found our original design discussion with Pieter - https://github.com/zeromq/zproject/pull/373

PR #826, opened 2017-01-09
Problem: make distcheck does not see it is under .git repo

jimklimov:
Solution: beside checking under srcdir/.git, also ask GIT if it can find the workpace up the tree

Follow-up to discussion in #820 : now `make all && make distcheck` is consistent regarding enablement of Draft API by default (without explicit flags) when running somewhere under a git repo.


bluca:
Thanks! Could you please do the same for CMake?

PR #828, opened 2017-01-09
Problem: want cmake to find git workspace root in higher level directories

jimklimov:
Solution: there is always time for a first attempt to write CMake syntax code ;)

Sibling to #826 

Note: testing with a real cmake project needed, no idea if I got the syntax to compare result code right

### DIFF
--- a/zproject_cmake.gsl
+++ b/zproject_cmake.gsl
@@ -40,7 +40,21 @@ if (NOT CMAKE_BUILD_TYPE)
     if (EXISTS "${SOURCE_DIR}/.git")
         set (CMAKE_BUILD_TYPE Debug)
     else ()
-        set (CMAKE_BUILD_TYPE Release)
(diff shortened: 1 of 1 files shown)


zproject_cmake.gsl
```diff
@@ -40,7 +40,20 @@ if (NOT CMAKE_BUILD_TYPE)
...
+            WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+            RESULT_VARIABLE git_result
+            OUTPUT_VARIABLE git_root
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+        )
+        message(STATUS "git workspace root [${git_result}]: ${git_root}")
+        if ( "${git_result}" == "0" )
+            set (CMAKE_BUILD_TYPE Debug)
+        else ()
+            set (CMAKE_BUILD_TYPE Release)
```
bluca: Missing endif ()

jimklimov: Thanks, updated

PR #831, opened 2017-01-11
Problem: cfg.example magically added to debian/rh recipes

vyskocilm:
Solution: drop that magic


bluca:
Any reason not to package the example config? Many packages do it

vyskocilm:
Well, they are packaged, but not installed. There used to be quirk for
make, but it was removed.

I'll try to extend zinstall model to support this.

Dne 11. 1. 2017 9:43 napsal uživatel "Luca Boccassi" <
notifications@github.com>:

PR #833, opened 2017-01-12
do not use specific locations for executables

bbdb68:
for windows build, the .exe should not be separated from executables. this makes interactive debugging in visual studio easy, and allows to fix the execution of unit test on appveyor, see https://github.com/zeromq/czmq/pull/1601/

I do not know if this change should be msvc only or could be applied for all platforms

(BTW is cmake used only on windows ?)

### DIFF
--- a/zproject_cmake.gsl
+++ b/zproject_cmake.gsl
@@ -231,9 +231,6 @@ add_library($(project.linkname) ${$(project.linkname)_sources})
 set_target_properties($(project.linkname)
     PROPERTIES DEFINE_SYMBOL "$(PROJECT.PREFIX)_EXPORTS"
 )
-set_target_properties($(project.linkname)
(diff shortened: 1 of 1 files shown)


bluca:
Cmake is used on OS x and Linux too, later I'll double check that there wasn't any side effect. Thank you!

jimklimov:
CMake works also on Solaris and illumos spinoffs, FWIW. But I'm not really convinced it brings substantial benefits compared to autotools - they seem to play in the same niche and solve same problems (and add different new ones to wrap a head around). IMHO autotools are more flexible and configurable, at least as far as a maintainer of distro management recipes is concerned ;)

bluca:
I'm not a huge fan of cmake myself, a lot of things it does are questionable at best (like overwriting cppflags).
But apart from Windows, it's necessary for our libraries so that projects that use cmake can include them easily

Issue #834, opened 2017-01-12
Problem: zinstall might not be the solution


sappo:
Continuing the discussion from #825.

@vyskocilm thanks for the thorough explanation!

@vyskocilm @bluca @jimklimov Here's what I think about the current solution:
* zinstall is not installing anything but providing snippets for other files!
* There is no need to store those snippets in actual files in each project. This just feels weird.

Here's what I propose:
1. Move everything concerning systemd into it's own target, i.e. `zproject_systemd`.
2. Allow targets like autotools to register a snippet hook for its various files.
3. Now the systemd target will provide the autotools target with snippets which are inserted during code generation and thus not stored on each user's project.

As a result the systemd stuff will vanish from the autotools, debian and redhat gsl files making them more readable and targets that scatter their stuff all over the place can be contained into one file.

Questions!? Remarks!?

jimklimov:
Such modularity makes sense to me - I think for autotools (and configure script) the reasonable approach would be to populate relevant files into the `projectroot/m4` directory and referencing the file by a single command in `configure.ac`. And of course pre-generate (the .in templates for) actual configs, units, includable pieces of the Makefile to set up install location and dist/clean requirements, etc.

Just as well, the common bit in the configure script might gain a `--with-user=...` and `--with-group=...` arguments (defaulted to `nobody`?) so a distro maintainer can specify which credentials the built daemons/services will run in his OS.

This could be non-invasively (readability!) generalized for other frameworks, like init-scripts for older distributions and SMF manifests and methods for Solarish systems, for example.

vyskocilm:
OK I see your main complaint now! You don't like the fact zproject_install and rest of the zproject scripts communicates using files :) All it needs is someone with much better gsl skills than I have.

PR #835, opened 2017-01-12
Problem: want to install additional config files

vyskocilm:
Solution: add type="config" to zproject_install.gsl

### DIFF
--- a/zproject_install.gsl
+++ b/zproject_install.gsl
@@ -22,6 +22,7 @@ function zinstall_include (filename)
     endif
 endfunction
 
+directory.create ("builds/zinstall")
 output "builds/zinstall/configure.ac"
 >
 output "builds/zinstall/Makemodule.am"
@@ -31,7 +32,6 @@ output "builds/zinstall/$(project.name).spec"
 output "builds/zinstall/$(project.name).install"
 >
 
-directory.create ("builds/zinstall")
 for project.main where defined (main->install)
     for install
 
--- a/zproject_install.gsl
+++ b/zproject_install.gsl
@@ -34,6 +34,7 @@ output "builds/zinstall/$(project.name).install"
 directory.create ("builds/zinstall")
 for project.main where defined (main->install)
     for install
+
         if install.type = "systemd-tmpfiles"
 
             append "builds/zinstall/configure.ac"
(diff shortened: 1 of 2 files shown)


zproject_install.gsl
```diff
@@ -22,6 +22,7 @@ function zinstall_include (filename)
     endif
 endfunction
 
+directory.create ("builds/zinstall")
```
sappo: Should only be created if zinstall is used!

sappo:
I'm not sure this problem is valid. Configuration files should never be part of the project itself! What are those configuration files?

sappo:
@vyskocilm do you have an example project?

zproject_install.gsl
vyskocilm: I tried the simplest way how to do it
project.has_install = (count (main, (count (main.install) > 0)) > 0)

but it crashes gsl. I'll create minimal reproducer and add issue for imatix/gsl.

Can we solve it in another PR, please?

vyskocilm:
@sappo https://github.com/42ity/fty-metric-composite/blob/master/src/fty-metric-composite.cfg.example

The example configuration is treated as a template for various instances of other services. Basically what can be changed is to drop type="config" and create a type="datadir". Makes a sense for you?

jimklimov:
I guess this installation of configuration files refers to sample configs
and/or those generated from templates (like autotools .in support) during
build for a particular distribution with particular FS layout (if that is
involved/specified in the config file), user account names etc. - things
that are not hardcoded in project sources and imposed on all distros, but
that can well be automated to simplify secure setup and working out of the
box on the final system.


2017-01-12 11:01 GMT+01:00 Kevin Sapper <notifications@github.com>:

zproject_install.gsl
sappo: A simple solution might be:
```
for project.main where defined (main->install)
    if !file.exists ("builds/zinstall")
        directory.create ("builds/zinstall")
    endif
...
```

sappo:
I understand the problem, still I think a user wont get the model. But lacking any better ideas I'll merge it for now.

sappo:
@vyskocilm #834 might be a cleaner solution.

Issue #567, opened 2016-03-02
Problem: --without-xyz does not work


hintjens:
I was wondering why `--without-uuid` refused to work in CZMQ. Finally I ripped the dependency out, as it's entirely pointless (to generate 32 random bytes does _not_ need an external, non-portable, eternally different library). 

However, note this code in `zproject_autotools.gsl`:

```
        AC_ARG_WITH([$(use.libname)],
            [
                AS_HELP_STRING([--with-$(use.libname)],
                [Specify $(use.libname) prefix])
            ],
            [search_$(use.libname)="yes"],
            [])
```

Now, the `--with` option in autoconf is conventionally used to select/deselect an optional package. It is _not_ used to specify the prefix for a package.

We should not be creating surprises for the user.

jimklimov:
I'm digging in this area of code for the past few days, and it seems that the de-facto syntax is to indeed enable/disable search for an optional package. However the argument can also take a value (beside implicit "yes" or the currently ignored "no"), which is then treated as the path such as a source code build workspace, under which there are an `include/` and `lib/` subdirectories containing the necessary files of this dependency.

I'm not triaging several ways to reorganize this code, hopefully will come up with a PR soon.

jimklimov:
With PR #838 in place, the `--without-stuff` works as well as `--with-stuff=no`, so the specific scope of this issue is solved.

PR #838, opened 2017-01-14
Steps to solve Issue #567

jimklimov:
And these are just steps on a path to solving a larger problem ;)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -68,6 +68,21 @@ else
     AM_CONDITIONAL(WITH_GCOV, false)
 fi
 
+# Set pkgconfigdir
+AC_ARG_WITH([pkgconfigdir], AS_HELP_STRING([--with-pkgconfigdir=PATH],
+    [Path to the pkgconfig directory [[LIBDIR/pkgconfig]]]),
+    [pkgconfigdir="$withval"], [pkgconfigdir='${libdir}/pkgconfig'])
+AC_SUBST([pkgconfigdir])
+
+# Use the provided pkgconfigdir not only to write our pkg-config data,
+# but also as an additional location to look for other packages metadata.
+AS_IF([test -n "${pkgconfigdir}" -a -d "${pkgconfigdir}"],
+    [AS_IF([test -z "${PKG_CONFIG_PATH}"],
+        [PKG_CONFIG_PATH="${pkgconfigdir}"],
+        [PKG_CONFIG_PATH="${pkgconfigdir}:${PKG_CONFIG_PATH}"])
+     export PKG_CONFIG_PATH
+    ])
+
 # Will be used to add flags to pkg-config useful when apps want to statically link
 PKGCFG_LIBS_PRIVATE=""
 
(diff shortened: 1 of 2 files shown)


jimklimov:
Tested for `--with-depname=no` and `--without-depname` syntax, as well as for `--with-depname` to enable an optional component (now off by default). 

Did not quite succeed with testing `--with-depname=path` because an earlier pkgconfig test wins. This may be or not be a problem to solve, but a different one.

jimklimov:
Review of the code in this area uncovered a few uncertainties, marked with TODO comments rather than logged issues (because I did not prove or disprove they are issues in reality, but if this does explode for someone - comments can help get them started fixing it).

sappo:
@jimklimov great job tackling this issue! But with all the compatibility layers you introduced the configure file is no longer maintainable. IMO we should simplify it to most obvious features the user will expect even if that means killing backwards comparability. I.e. use with/without switches to enable/disable usage of a library/program and don't pass any yes/no or path values here!

@bluca  @vyskocilm @karolhrdina what do you think?

bluca:
Sorry, haven't had time to look in details. Will use the new setup next week and see how it looks like on our various projects.

In general I agree that simple is better and things need to be maintainable especially as new developers come onboard, but on the other hand supporting the de-facto *NIX standard parameters and options is a big win for new users of the generated projects.

jimklimov:
The automake `with` feature is interpreted by the standard `configure` engine (during script startup) and has its own semantics. A `with-something` without equality to a value is assigned a `withval_...` equal to string `yes`. A `without-something` is identical to `with....=no`. And any other custom value is assigned into the `withval_...` - such as the paths to a custom source/build subtree, in this use-case.

Note that in the original code, there was no enable/disable influence from this switch - it was (non-standardly) used to pass the path to custom directory, and that codepath only worked (and still does) only if the pkg-config data for that project was not discovered earlier. I think this may be also a problem (if I explicitly ask to build against this custom code, why does it still build against a standard package that is also present), but if accepted - it is a problem for another PR to tackle :)

With this PR in place, the ability to specify a custom path is retained (and makes sense, especially if re-made to take priority when specified), but also we gain an ability to disable searching for certain dependencies. For a slightly stretched example, I as an OS maintainer understand that the program will not work without say malamute, but I don't have it packaged on my build host... and in fact, the job in question only wants to build manpages for the distro docs site. In this semipractical case, I can configure `--without-libmlm`, the binaries won't build, but I only want to `make doc` so I don't care.

sappo:
I guess it makes sense but you moved deep into dark pits of the daemon realm for me to fully grasp what you tried to explain :D

jimklimov:
Well, this is a twilight realm :)

Application developers (unless they are from devops bunch) generally do not and perhaps should not concern themselves with how the results of their coding would be deployed - likely they won't even know in advance on what OSes this ends up running. So the code is written (or wrapped) as portable as possible.

However, the project files do also commonly include some aid for OS distro packagers, who prefer their least-effort object-oriented approach "build me a package from this tarball" which standard frameworks like autotools (or I guess CMake) allow to fulfill. When you package tens of different softwares a day, you do not really want to dig into nuances of each - you want them to look and behave standard to whatever extent possible. Like with initscripts (or services) - you have a call to start and stop something, and as long as it works - you do not care much for the magic that happens behind those entry-points.

And when it stops working... then it often becomes my concern ;p

vyskocilm:
Hi kevin,

Sorry, my automagic skills are light years behind jim's, so I must say I am
lost.

I was prototyping a list of aletrnate pkg config names, but never got so
far. Plus jim have invested more effort. Sadly to see see that lua is so
big mess made by our fellow package maintainers.

I share your concerns about the complexity ... But well, maybe it will push
me in my quest of having simpler build back-end (gyp or premake, ninja or
whatever) in zproject :-)

Nice weekend to all of you!

Dne 15. 1. 2017 15:41 napsal uživatel "Kevin Sapper" <
notifications@github.com>:

PR #842, opened 2017-01-19
Problem: class type definitions can't be used by other classes

bkmit:
Solution: move forward class type defintions from class include file to _classes.h file.

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -418,10 +418,17 @@ $(project.GENERATED_WARNING_HEADER:)
 //  External API
 #include "../include/$(project.header:)"
 
-//  Internal API
+//  Extra headers
 .for header where scope = "private"
 #include "$(header.name).h"
 .endfor
+
+//  Opaque class structures to allow forward references
+.for class where scope = "private"
+typedef struct _$(class.c_name)_t $(class.c_name)_t;
+.endfor
+
+//  Internal API
 .for class where scope = "private"
 #include "$(class.c_name).h"
 .endfor
--- a/zproject_skeletons.gsl
+++ b/zproject_skeletons.gsl
@@ -343,10 +343,6 @@ extern "C" {
 #include "$(class.c_name)_library.h"
 
 .   endif
-.   if class.scope = "private"
-typedef struct _$(class.c_name)_t $(class.c_name)_t;
-
-.   endif
 //  @interface
 .   if class.scope = "private"
 .      visibility = "$(PROJECT.PREFIX)_PRIVATE"


bluca:
@bkmit this breaks the build in CZMQ:

```
In file included from src/czmq_classes.h:33:0,
                 from src/zactor.c:34:
src/zgossip_msg.h:21:31: error: redefinition of typedef 'zgossip_msg_t' [-Werror=pedantic]
 typedef struct _zgossip_msg_t zgossip_msg_t;
                               ^
In file included from src/zactor.c:34:0:
src/czmq_classes.h:30:31: note: previous declaration of 'zgossip_msg_t' was here
 typedef struct _zgossip_msg_t zgossip_msg_t;
```

Could you please have a look?

bkmit:
I'm looking to this issue right now.

bkmit:
Look like only part of class .h file between @interface and @end delimiters get regenerated if file exists. Unfortunately class typedef forward definition is outside this region.
Simple removing of zgossip_msg.h before gsl project.xml fixes this bug.
Hope this help.

bluca:
Internal definitions were already usable before by including $project_classes.h are you sure you weren't simply missing that in your project?

bkmit:
Yes, it is useable, but only in .c files. I can't use types of one class for definition argument and return in $class.api files of other classes. Such definitions are propagated to $class.h file and are undefined at compile time. My patch fixes this.

bluca:
I don't follow. The typedef moved from the class header to the general header. But immediately after defining the typedef, the private class headers are included. So apart from breaking the build, it also doesn't make any difference - if the typedef can be used now, then it could be used before because of the include, and indeed it is in CZMQ for example

bluca:
@bkmit ping :-)

bkmit:
@bluca pong :-), Sorry for later response, missed your message.
Yes, it can be used by C code. But if *type* defined by this typedef used in (**pravate**) $class.api file for defining method argument or return *type* gsl will produce $class.h where this *type* used.
$class.h files are included by $project_classes.h in order of definition in $project.xml which lead to error **'type undefined'** at compile time. In contrast **public** class typedef's already moved to $project_library.h file! My PR do the same for **private** $class. I have no good idea how to properly fix build breakage other than removing $class.h files before regeneration of project.

bluca:
Can you link an example?

bluca:
@bkmit I think I see what you mean now. I'll fix up manually all our projects

bkmit:
@bluca Thank you for patience!

bluca:
No problem, thanks for explaining!

I noticed there is also a compatibility problem with code generated by zproto.
zproto wraps the forward declarations in ifdef guards, it's probably best to add it here too to avoid more issues, I'll take care of it

PR #848, opened 2017-01-23
Do not use globs for systemd unit files

jimklimov:
Following up on #811 - it seems that debian always and RPM sometimes do not well support globs. So another solution is proposed to better handle the single-instance and/or multi-instance service units. Non-service units are left for the future, so far no released zproject version shipped them out of the box.


bluca:
thanks!

bluca:
I think on Debian dh-exec might have a better support for wildcards

PR #859, opened 2017-02-16
Problem: distcheck option in README.md is wrong

vyskocilm:
Solution: fix it accroding gsl script


vyskocilm:
I was wrong, distcheck is the correct name

PR #860, opened 2017-02-23
Problem: difficult to change Makemodule-local.am support

jimklimov:
WAS:Problem: zproject regeneration required to add (or remove) src/Makemodule-local.am support

Solution: always generate the line with "-include" to pull this file if it exists at build-time (the minus-cmd syntax may be a GNU Make extension though). Also fix a typo in comments.

Bonus question to consider *before merging* : does use of zproject require use of GNU Make (do we care about possibly breaking other Make's)? Per docs,

https://www.gnu.org/software/make/manual/html_node/Include.html
> For compatibility with some other make implementations, sinclude is another name for -include.

Should we bother to use this keyword instead?


bluca:
If there's an alternative that works on multiple make implementations then I'd say go for it

jimklimov:
Ok, changed. Googling says this should work in BSD, Sun, SGI makes too (although "-include" seems to have come from BSD make - which is a baseline for GNU make). Also GMake docs confirm it in https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_12.html

PR #861, opened 2017-02-23
Problem: some generated sources place internal headers into src/

jimklimov:
Solution: add -Isrc/ to CPPFLAGS; just in case for build-time generated headers (e.g. from templates) also add the builddir/include and builddir/src.

NOTE: This is about internal headers (implementation details of classes and APIs) as far as I've seen, so it makes little sense to expose them in include/ dir properly. But feel free to decide more authoritatively if any of those should be generated into another path; out of scope for this PR though :)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -10,7 +10,8 @@ AM_CFLAGS = \
 
 
 AM_CPPFLAGS = \
-    -I$(srcdir)/include
+    -I$(srcdir)/include -I$(srcdir)/src \
+    -I$(builddir)/include -I$(builddir)/src
 
 project_libs =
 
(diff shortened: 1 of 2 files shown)


bluca:
I think this is fine, we don't want internal headers in include/

jimklimov:
As @karolhrdina noted, this PR did not propose proof of an existing problem that this commit solves.
So far I've quickly come up with one of our project repos where distcheck fails without the added include-dir flag, but the legacy of the header file in that case is not apparent (might be not generated by zproject at all, or not by its current version).

So *possibly* this PR avoids not an issue that zproject itself generates into projects (or we have to dig deeper to coerce it into generating an inconsistent project), but one that adventurous developers with private headers can - and do/likely-would - burden unto themselves. Thus, it still does solve a practical problem :)

PR #863, opened 2017-02-27
Revert "Problem: some generated sources place internal headers into src/"

karolhrdina:
Reverts zeromq/zproject#861

Problem statement by @jimklimov is **not a problem**:
1. Internal headers *should* be placed into src/
2. As @bluca mentioned in the [comment](https://github.com/zeromq/zproject/pull/861#issuecomment-282019544), internal headers *should not* be placed into /include,  but that is **NOT** happening. There is nothing to be fixed!
3. No matter what, a cleanly generated project (i.e. no manual wiring) will never fail `make distcheck`.

The real problem @jimklimov was trying to fix with this PR was that one of our repos was failing `make distcheck` on his system. The reason why this repo failed is not because of zproject but because someone made manual changes to the wiring/includes. 

@jimklimov: If you provide a project.xml that would, after generation (without faulty manual additions), fail make distcheck on any system i am more then willing to re-visit this topic.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -10,8 +10,7 @@ AM_CFLAGS = \
 
 
 AM_CPPFLAGS = \
-    -I$(srcdir)/include -I$(srcdir)/src \
-    -I$(builddir)/include -I$(builddir)/src
+    -I$(srcdir)/include
 
 project_libs =
 
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -948,8 +948,7 @@ AM_CPPFLAGS = \\
 .if project.use_cxx
     -D__STDC_FORMAT_MACROS \\
 .endif
-    -I\$(srcdir)/include -I\$(srcdir)/src \\
-    -I\$(builddir)/include -I\$(builddir)/src
+    -I\$(srcdir)/include
 
 .libs = ""
 .for use where use.libname ?<> ""


karolhrdina:
I type faster then i think - Addendum #1: I forgot to write that the problem @jimklimov is having has to be fixed in the appropriate repo  and not in zproject. The private include (draft classes) are hidden by design!

jimklimov:
The curious case of... (digging for reasons, but popped up anyhow):

````
g++ -I/var/scratch/tmp/FTY-tmp1/.install/Linux-x86_64/usr/include -pedantic -Werror -Wall -D_GNU_SOURCE -DLINUX -DFTY_ALERT_ENGINE_INTERNAL_BUILD -D_REENTRANT -D_THREAD_SAFE  -I/var/scratch/tmp/FTY-tmp1/.install/Linux-x86_64/usr/include -L/var/scratch/tmp/FTY-tmp1/.install/Linux-x86_64/usr/lib  /var/scratch/tmp/FTY-tmp1/fty-alert-engine/src/fty_alert_engine_selftest.cc /var/scratch/tmp/FTY-tmp1/.build/Linux-x86_64/fty-alert-engine/src/selftest-rw /var/scratch/tmp/FTY-tmp1/.build/Linux-x86_64/fty-alert-engine/src/selftest-ro   -o src/fty_alert_engine_selftest
In file included from /var/scratch/tmp/FTY-tmp1/fty-alert-engine/src/fty_alert_engine_selftest.cc:30:0:
/var/scratch/tmp/FTY-tmp1/fty-alert-engine/src/fty_alert_engine_classes.h:30:22: fatal error: platform.h: No such file or directory
 #include "platform.h"
                      ^
compilation terminated.
<builtin>: recipe for target 'src/fty_alert_engine_selftest' failed
make[2]: *** [src/fty_alert_engine_selftest] Error 1

jim@debian8:/var/scratch/tmp/FTY-tmp1$ ls -al /var/scratch/tmp/FTY-tmp1/.build/Linux-x86_64/fty-alert-engine/src/platform.h 
-rw-r--r-- 1 jim jim 6720 Feb 28 18:13 /var/scratch/tmp/FTY-tmp1/.build/Linux-x86_64/fty-alert-engine/src/platform.h
```

Note that this is an out-of-tree build, so the `platform.h` generated from the template is in the builddir, while the `fty_alert_engine_classes.h` which includes it (with double-quotes) is in the original checked-out workspace.

I'm looking at what could cause this (possibly other experiments with zproject features I have in development), but certainly the include paths are needed.

So currently consider this as anecdotal evidence that may later prove irrelevant...

PR #864, opened 2017-02-28
Problem: Only main is considered for systemd

aquette:
Solution: Add snippets for bin programs too

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,3 +19,4 @@ Dave Meehan
 Wes Young
 Kouhei Sutou
 Andreas Potschka
+Arnaud Quette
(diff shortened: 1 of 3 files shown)


sappo:
@aquette thanks for your PR. We lately started to group stuff like systemd which scatters all of its pieces across multiple files into one gsl file. Which for systemd this is `zproject_systemd`. Could you please move your snippets into this file. Just copy them to the functions `function snippet_redhat_spec_systemd` and `function snippet_debian_spec_systemd` and the result should be the same as now.

aquette:
@sappo I'd gladly do it, but my zproject understanding is currently too limited, as my time, to push something optimal for both main and bin, along with .service and .timer into zproject_systemd, without duplication.
I've made an issue to track that, which is the best I can do for now...

PR #865, opened 2017-02-28
Problem: project-local CMake hook are not included

aquette:
Solution: Fix typo error

### DIFF
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -1003,7 +1003,7 @@ EXTRA_DIST += \\
     src/$(project.prefix)_classes.h
 
 include \$(srcdir)/src/Makemodule.am
-sinclude \$(srcdir)/src/Makemodule-local.am # Optional project-local hook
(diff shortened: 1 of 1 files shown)


jimklimov:
Not a typo, `sinclude` is the silent-include keyword (see recent PR #860 that added it). AKA the `-include` in GNU make, to try including the file but not bomb out if it is missing (so we do not have to regenerate the zproject after adding the manually maintained file to codebase).

Namely, this change-back will break projects that do not have an `src/Makemodule-local.am` ;)

jimklimov:
@sappo : you merged a not-yet-green PR (and it would end up red because zproject's Makefile.am was not amended to match the "fixed" template). That's beside not thinking about what you merge ;)

jimklimov:
@aquette : does the change you proposed fix something on your builds (e.g. did `sinclude` break anything with a particular `make` implementation)?

aquette:
@jimklimov sorry, I indeed misunderstood the behavior and wrongly thought that this should be expanded in the final Makefile, while it's indeed a runtime expansion... Reverting commit

PR #866, opened 2017-02-28
Revert "Problem: project-local CMake hook are not included"

jimklimov:
Reverts zeromq/zproject#865 because it will break downstream projects (and is not fixing an error in the first place). See also #860.


sappo:
@jimklimov thanks!

PR #869, opened 2017-02-28
Problem: projects stopped EXTRA_DISTing Makemodule-local.am after PR#860

jimklimov:
Solution: detect it at configure time and redist optionally

Thanks to @aquette for persisting that PR #860 did introduce a bug.


aquette:
thanks @jimklimov , your PR should indeed fix this, though I've not tested (only PR review)

PR #871, opened 2017-02-28
Problem: maintenance of selftests (ab)using files and distcheck is chaotic

jimklimov:
Solution: generalize the approach used in our projects, so this ritual can be maintained by zproject without need for custom Makemodule-local.am at least for this use-case (note: someone else should ditto this into CMake recipes).


jimklimov:
WIP for now, testing locally that it does not break builds

jimklimov:
Now tested by simplifying another one of our projects using this infrastructure.

src/Makemodule.am
```diff
@@ -53,36 +53,84 @@ dist_bin_SCRIPTS = \
...
+
+CLEANFILES += $(abs_top_builddir)/$(SELFTEST_DIR_RW)/*
+
+# Note that this syntax dists the whole directory - including subdirs (if any)
+EXTRA_DIST += $(SELFTEST_DIR_RO)
+
+clean-local: clean-local-selftest-ro clean-local-selftest-rw
+.PHONY: clean-local-selftest-ro
+clean-local-selftest-ro:
+	@if test "$(abs_top_builddir)" != "$(abs_top_srcdir)" ; then \
```
sappo: @jimklimov what does the `abs` stand for?

jimklimov: "abs"olute paths, as resolved on the host during build time
"top" is for the highest-level (first-called) Makefile location, e.g. project workspace root
"src" is where sources lie, "build" is where build runs - may be a completely different filesystem or just a subdirectory of this one (as used by distcheck and friends).

Default ones are relative (to some makefile's directory, and/or caller's CWD. In situations like this we explicitly want to compare full paths to make sure they are different.

PR #873, opened 2017-03-01
Problem: Improvements in PR#864 did not generate files from templates

jimklimov:
Solutions:
* generate configure.ac snippets for systemd service and timer .in templates, where available;
* add gitignores for such generated files;
* support systemd timer units (if sources are present) not only for BINs but also MAINs to be on par (redhat, debian)

Follow-up for PR #864 ...


jimklimov:
Thanks for merge, and for initial diagnosis to fix this first :)

PR #877, opened 2017-03-01
Problem: Generated Makefile had strange rules for src: and code:

jimklimov:
Solution: My guess at how to fix it, real review is desirable :)

@sappo : you seem to have latest git-blame hits in this area, plz take a look before just merging :)

### DIFF
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -1323,11 +1323,11 @@ dist_api_DATA = \\
 .endfor
 .if count (project.class)
 # define custom target for all products of /src
-src:
+src: \\
 .   for project.main
-\tsrc/$(main.name) \\
+\t\tsrc/$(main.name) \\
 .   endfor
-\tsrc/$(project.libname).la
+\t\tsrc/$(project.libname).la
 
 .endif
 .for model
@@ -1348,8 +1348,10 @@ code:
 .       endif
 -q $(name).xml
 .   endif
-.endfor
+.   if last ()
 \tcd $\(srcdir); gsl -target:- project.xml
+.   endif
+.endfor
 
 # Directories with test fixtures optionally provided by the project,
 # and with volatile RW data possibly created by a selftest program.
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ config.log
 config.status
 config/
 configure
+configure.lineno
 libtool
 src/platform.h
 src/platform.h.in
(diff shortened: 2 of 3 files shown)


jimklimov:
Diff of generated codes : 

````
 # define custom target for all products of /src
-src:
-       src/fty-metric-store \
-       src/fty_metric_store_selftest \
-       src/libfty_metric_store.la
-
-       cd $(srcdir); gsl -target:- project.xml
-
+src: \
+               src/fty-metric-store \
+               src/fty_metric_store_selftest \
+               src/libfty_metric_store.la
````

So now `src` depends on those listed binaries and libs (rather than trying to run them - which was probably not intended), and there is no dangling `gsl` call outside any recipe...

File: zproject_autotools.gsl [hunk 1]
```diff
@@ -1323,11 +1323,11 @@ dist_api_DATA = \\
 .endfor
 .if count (project.class)
 # define custom target for all products of /src
-src:
+src: \\
```
sappo: This looks wrong. Correct me if I'm wrong but make targets don't continue on next line.

zproject_autotools.gsl
```diff
@@ -1323,11 +1323,11 @@ dist_api_DATA = \\
...
 .   for project.main
-\tsrc/$(main.name) \\
+\t\tsrc/$(main.name) \\
 .   endfor
-\tsrc/$(project.libname).la
+\t\tsrc/$(project.libname).la
```
sappo: Why add a second tab. This will kill the formatting.

zproject_autotools.gsl
jimklimov: You can concatenate long lines, at least in many `make`s... Can probably make it generate a single-line in GSL though.

The code generated earlier defined a target called `src` with no dependencies and whose action was to run a single command made up of names of binaries and the `.la` lib. Now, THAT certainly seemed wrong :)

zproject_autotools.gsl
jimklimov: I thought on the opposite - at least initially, where it was hard to discern the (presumed) dependencies from the recipe rules (the `gsl` call below).

Anyhow, if we concatenate the `src` deps into one line, this nit will disappear too :)

zproject_autotools.gsl
sappo: Yes you're right, just checked this command in zyre. It doesn't do anything.

PR #880, opened 2017-03-01
Problem: "include" used is not from Make but from Autotools

jimklimov:
Solution: if we want to always be ready to include the Makemodule-local.am, we have to ensure one always exists...


sappo:
Thanks!
Btw. is just me or did Github change their color palette?

jimklimov:
Seems the purple merges were darker yesterday...

BTW: tested this by regenerating czmq - basic make stuff passed, distcheck in progress...

PR #883, opened 2017-03-06
Problem: Debian packaging misses src/Makemodule-local.am if not in SCM

jimklimov:
Solution: call autogen.sh before building to generate the dummy

Follow-up to #860 et al

Improvement tested in our project - passed public Travis and private Jenkins + OBS

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -25,16 +25,23 @@ endif
 # Workaround for automake < 1.14 bug
 $(shell dpkg --compare-versions `dpkg-query -W -f='$${Version}\n' automake` lt 1:1.14 && mkdir -p config)
 
+.autogened: autogen.sh
+	@echo "Executing autogen.sh for this project..."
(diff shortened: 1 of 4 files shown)


packaging/debian/rules
```diff
@@ -25,16 +25,23 @@ endif
...
 override_dh_auto_test:
 	echo "Skipped for now"
 
-override_dh_auto_configure:
+override_dh_auto_configure: .autogened
 	dh_auto_configure -- \
 		--enable-drafts=$(DRAFTS)
 
 %:
+	if test "$@" = "clean" ; then rm -f .autogened ; fi; \
```
bluca: better to override dh_auto_build and dh_auto_clean

bluca:
@jimklimov this breaks the package build:

https://build.opensuse.org/package/live_build_log/network:messaging:zeromq:git-stable/zproject/Debian_9.0/x86_64

PR #885, opened 2017-03-07
Problem: debian/rules "autogened" should not call extra $(MAKE)

jimklimov:
Solution: define makefile-style targets to call or clean-up the autogen script within currently processed set of recipes rather than calling a $(MAKE) with an undefined makefile name ;)

Follow-up for #883

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -30,6 +30,9 @@ $(shell dpkg --compare-versions `dpkg-query -W -f='$${Version}\n' automake` lt 1
 	./autogen.sh
 	touch $@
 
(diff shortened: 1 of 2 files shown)


bluca:
Why not simply override dh_auto_clean and dh_auto_build?

jimklimov:
On one hand, I don't know the DH rules that well so I fit into what exists. On the other, I know that these rule names do the job for our OBS, so I fit into what exists. On the third hand (oops) I don't quite want to repeat the rules defined by two pages of GSL for the `%` case, so... you get the picture :)

PR #890, opened 2017-03-28
Problem: Generation of systemd service is missing

aquette:
Solution: Add some complementary information

When using both autotools and systemd, since configure.ac is generated first
and checks for the actual presence of service file to list these in
AC_CONFIG_FILES directive, and since the generation of these service files is
done after the configure.ac generation, it results that systemd service files
generation are missing from configure.ac, and thus the files are not generated
from the template. A workaround is to call twice `gsl project.xml`, but it is
worth a fix.


aquette:
Additional comment to reproduce (from a zproject generated project) which use autotools and systemd:

- rm src/*.service*
- rm configure.ac 
- First run: gsl project.xml
  result excerpt from configure.ac:
    ...
    AM_COND_IF([WITH_SYSTEMD_UNITS],
        [AC_CONFIG_FILES([
        ])],
        [])

- Second run: gsl project.xml
  result excerpt from configure.ac:
    ...
    AM_COND_IF([WITH_SYSTEMD_UNITS],
        [AC_CONFIG_FILES([
                     src/<some-service-file>.service
        ])],
        [])

PR #893, opened 2017-04-06
Problem: current integration with ASAN can break CFLAGS

jimklimov:
Solution: remake by producing a custom configure option

Follow-up on #892 

In any case, having this automated can be beneficial for not-only-travis test builds.

PR #894, opened 2017-04-06
Configure script: the flag type for ASAN is "enable", not "with"

jimklimov:
Also add messages to make ASAN and GCOV enablement more visible

Follow-up to #892 and #893

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -53,6 +53,7 @@ AC_PROG_AWK
 PKG_PROG_PKG_CONFIG
 
 # Code coverage
+AC_MSG_CHECKING([whether to enable GCov])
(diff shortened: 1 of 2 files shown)


jimklimov:
@bluca : can you test the resulting builds (that they actually do compile and test-sanitize the code)? IIRC there may be more build/link options related to this, e.g. https://gist.github.com/kwk/4171e37f4bcdf7705329

Maybe it makes sense to follow up with a `make` target similar to what is done for valgrind and gcov, adding relevant envvar settings, etc..

bluca:
looks fine, thanks:

https://travis-ci.org/bluca/czmq/builds/219275488

The flag seems to be enough - I've used it before and it found issues in other projects

PR #898, opened 2017-04-10
Problem: generated .travis.yml repeats too much

jimklimov:
Solution: DRY is good. Don't WET yourself (especially if manually). Use YAML references to define and maintain just once the list of dependency packages and their repos (e.g. OBSed ones from zeromq ecosystem).

Also made default builds docless (so no overhead of installing doc generation packages as well as running them), while a build with docs is handled separately - and earlier in the matrix chain, so overall the build does not wait for it alone to finish.

Example victim file: https://github.com/jimklimov/fty-proto/blob/test-yaml-refs/.travis.yml

### DIFF
--- a/zproject_travis.gsl
+++ b/zproject_travis.gsl
@@ -94,7 +94,6 @@ addons:
       <<: *pkg_src_zeromq_ubuntu12
     packages: &pkg_deps_common
     - *pkg_deps_devtools
-    - *pkg_deps_doctools
     - *pkg_deps_prereqs
 
 matrix:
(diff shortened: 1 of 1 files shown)


zproject_travis.gsl
```diff
@@ -44,79 +58,77 @@ env:
...
-        - $(string.replace (use.project, "_|-"))-dev
+    - $(string.replace (use.project, "_|-"))-dev
 .         endif
 .      endfor
+
+pkg_deps_doctools: &pkg_deps_doctools
+    - asciidoc
+    - xmlto
+
+pkg_deps_doctools: &pkg_deps_devtools
```
bluca: shouldn't this be:

```pkg_deps_devtools: &pkg_deps_devtools```

jimklimov: Oops, probably.  Although it is harmless - the name of the reference counts; name of this custom object is not used by anyone so it's effectively discarded.

PR #906, opened 2017-05-25
Problem: cffi bindings are slow to import and require C parser

vyskocilm:
Solution: reorganize a concept totally. Those are changes
 * create Python package $(project.name)_cffi
 * move definitions to one file
 * add dlopen module for older approach
 * add build.py to actually build extension module
 * let __init__.py try native and fallback to dlopen version

Tested on czmq_cffi bindings

### DIFF
--- a/zproject_python_cffi.gsl
+++ b/zproject_python_cffi.gsl
@@ -76,13 +76,23 @@ function resolve_class (class)
 endfunction
 
 function generate_binding
-    directory.create ("bindings/python_cffi")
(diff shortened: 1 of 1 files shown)


sappo:
I do have some major changes of the cffi binding laying around. I'll try to merge and push them on the weekend. Especially I solved the destructor problem!

sappo:
@vyskocilm I commit my changes to my fork: https://github.com/sappo/zproject/blob/master/zproject_python_cffi2.gsl

Issue #907, opened 2017-05-29
Missing target version


diorcety:
The SOVERSION is defined for the target but not the VERSION. Which imply that the libxxx.so.M.N.R is created but not libxxx.so.M. The python wrappers fail to find the libraries when compiling them with cmake

bluca:
Are you referring to cmake? Could you send a PR to fix it please?

vyskocilm:
Hi, our looks like fix had been merged by @bluca closed

PR #908, opened 2017-05-29
Add VERSION property and fix SOVERSION value

diorcety:
Related to #907 

From different sources it seems that SOVERSION should be only one number(abi?) and VERSION should be major.minor.patch

http://pusling.com/blog/?p=352
https://cmake.org/pipermail/cmake/2012-September/051902.html
https://gitlab.cern.ch/dss/eos/commit/18ff0746ff4bc1263648fe3fdda79075ce262093

### DIFF
--- a/zproject_cmake.gsl
+++ b/zproject_cmake.gsl
@@ -240,9 +240,19 @@ set_target_properties($(project.linkname)
 )
 set_target_properties ($(project.linkname)
 .if defined (project->abi)
-    PROPERTIES SOVERSION "$(project->abi.current - project->abi.age).$(project->abi.age).$(project->abi.revision)"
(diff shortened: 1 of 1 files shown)


bluca:
I'm not sure, it should do what libtool does to provide compatibility

bluca:
NVM, it does the right thing:

```
$ ls -l libczmq.so*
lrwxrwxrwx 1 lboccass lboccass     12 Jun 15 19:59 libczmq.so -> libczmq.so.4
lrwxrwxrwx 1 lboccass lboccass     16 Jun 15 19:59 libczmq.so.4 -> libczmq.so.4.0.3
-rwxr-xr-x 1 lboccass lboccass 906256 Jun 15 19:59 libczmq.so.4.0.3
```

Thanks for the fix!

Issue #909, opened 2017-06-02
jni singleton methods aren't static


jimrthy:
This shows up in 2 places right off the bat:
1. You can't call Zsock.newPull() to create a PULL socket (or any other variant on this theme).
2. You can't call Zframe.recv()

Instead, you have to create a new instance of some kind of socket that you might not care about in order to create the one you do. And then create a new instance of Zframe that you almost definitely don't want in order to recv messages.

jimrthy:
Is there any sort of community process for closing an issue that I wrote up and fixed myself? I should I just close this one because I think it's done?

bluca:
Yes, just mention the PR that fixed it and close the issue. Thanks!

jimrthy:
I think https://github.com/zeromq/zproject/issues/910 fixes it.

PR #910, opened 2017-06-02
Problem: JNI singleton methods aren't static

jimrthy:
Solution: Make the template functionality match the basic idea from
the python bindings.

This addresses https://github.com/zeromq/zproject/issues/909


jimrthy:
The generated code is https://github.com/zeromq/czmq/pull/1678

sappo:
Thanks!

PR #915, opened 2017-06-16
Problem: python_cffi bindings lacks standard build/install way

vyskocilm:
Solution: add setup.py, improve docs ...

sorry @sappo you've been too slow ;-)


sappo:
@vyskocilm Neat!

PR #917, opened 2017-06-19
Problem: I woked up early today

vyskocilm:
Solution: after a bath, relax, cleaning, ... I have finished a spec file generation for python_cffi bindings. This is a part of my 😈 plan to have 🐍 (cffi) the first class citizen for zproject based projects.

Works reasonably well, see
https://build.opensuse.org/package/show/home:mvyskocil:branches:network:messaging:zeromq:git-draft/czmq

The only remaining problem are systems without python3 interpreter. I need to find a solution for them.
And yes, Debian is not supported.

🖌 : and yes. Second problem right now generated bindings works only with DRAFT build ...

### DIFF
--- a/zproject_python_cffi.gsl
+++ b/zproject_python_cffi.gsl
@@ -185,6 +185,16 @@ function generate_binding
     >            else:
     >                ret [key].extend (value)
     >
+    >    # Python3 and strict unicode
+    >    for key, value in ret.items ():
+    >        if isinstance (value, bytes):
+    >            ret [key] = value.decode ("utf-8")
+    >        elif isinstance (value, list):
+    >            if isinstance (value[0], tuple):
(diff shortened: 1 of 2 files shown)


bluca:
Are the CFFI 1.0.0 and Python3 3.5 requirements both strictly necessary? Thinking about support for older distros

Also I'm adding the Debian support

vyskocilm:
Hi,

I don't think so. It should work on older versions.  Thanks for Debian part

Dne 20. 6. 2017 2:41 dopoledne napsal uživatel "Luca Boccassi" <
notifications@github.com>:

jimklimov:
It was nice of you guys to wake up so early, but you merged a travis-not-passed PR and left it so for half a day now. Booo :)

Also, this PR changes a seemingly unrelated matter about running `autogen.sh` only if one exists. Earlier the build would bomb-out if the file is not there. Are we sure we want/need this change (for all RPMs, but not other packages), or is it a leftover from experiments?

PR #920, opened 2017-06-21
Problem: build of czmq fails on platform w/o python3

vyskocilm:
SOlution: split python2/python3 bindings to separate spec file, so we can
better control who is going to use it

see https://build.opensuse.org/package/show/home:mvyskocil:branches:network:messaging:zeromq:git-draft

### DIFF
--- a/zproject_python_cffi.gsl
+++ b/zproject_python_cffi.gsl
@@ -190,6 +190,8 @@ function generate_binding
     >        if isinstance (value, bytes):
     >            ret [key] = value.decode ("utf-8")
     >        elif isinstance (value, list):
+    >            if len (value) == 0:
+    >                continue
     >            if isinstance (value[0], tuple):
     >                ret [key] = [(v[0].decode ("utf-8"), v[1].decode ("utf-8")) for v in value]
     >            else:
--- a/zproject_redhat.gsl
+++ b/zproject_redhat.gsl
@@ -42,16 +42,7 @@ register_target ("redhat", "packaging for RedHat")
 .   if !has_main & !project.exports_classes
 %global debug_package %{nil}
 .   endif
-.if python_cffi ?= 1
 
-# build with python_cffi support enabled
-%bcond_with python_cffi
-%if %{with python_cffi}
-%define py2_ver %(python2 -c "import sys; print ('%d.%d' % (sys.version_info.major, sys.version_info.minor))")
-%define py3_ver %(python3 -c "import sys; print ('%d.%d' % (sys.version_info.major, sys.version_info.minor))")
-%endif
-
-.endif
 Name:           $(project.name)
 Version:        $(project->version.major).$(project->version.minor).$(project->version.patch)
 Release:        1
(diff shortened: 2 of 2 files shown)


c-rack:
@vyskocilm following the link shows a 404 page

vyskocilm:
meh - it's not really REST style link :)

This one works
https://build.opensuse.org/project/show/home:mvyskocil:branches:network:messaging:zeromq:git-draft

bluca:
Having a separate spec will cause more than a few headaches I'm afraid, what about an additional bcond instead?

vyskocilm:
Can you elaborate it? I don't see a problem... For more bconds...  Can you
set it per repo in OBS?

Dne 21. 6. 2017 9:45 dopoledne napsal uživatel "Luca Boccassi" <
notifications@github.com>:

bluca:
It can, something like the following in the prjconf:

```
%if "%_repository" == "RHEL_7"
Macros:
%_with_python_cffi_3 1
%endif
```

As for the reasons, for starters there are limitations in the Github triggers: you can kick just ONE package, not multiple ones. So the second one cannot be kicked when we merge/push code in git.

Then there's the ABI issue - the cffi shared object would be built against a different library, since it's built separately - in the "main" spec the library is built and shipped in a package in the repository, and in the "python" spec the library is built and then thrown out, and only the python stuff is shipped in the repository.

Now this is not an issue for CZMQ since its build is fully reproducible given the same sources and environment, but there's no guarantee that it's always the case for all clients of zprojects, many of which we have no knowledge of.

Also a strict dependency, using rebuild counters, would also need to be added.

The simply having more than one spec/dsc per repository is just more work to maintain and build.

IMHO using a bcond is just easier and simpler.

bluca:
I can give it a try to implement later tonight or tomorrow - the Debian stuff is also ready, but I had issues with Ubuntu 14.04 which we still need to support for 2 years. Unfortunately python-cffi was in the "universe" repository and at that time OBS did not import those packages sadly (fortunately it does for newer releases), but it was a pain to build it locally since it was not happy at all with the sources from Debian Wheezy.

I managed to get it working last night:

https://build.opensuse.org/package/show/home:bluca:branches:network:messaging:zeromq:git-stable/czmq

I'll finish it up tonight and send a PR.

For the older RPM distros like RHEL/Centos 7 and SLE12 I tried as well but it's a dependency hell, in RHEL for example there's no Python 3 at all as you noticed so I fully agree with making the Python3 RPM optional.
I'll try to backport python2 CFFI for RHEL7.

Finally it turns out that Leap 42.2 discountinued i586, although the basic stuff is still built so we didn't notice until now, but half of the Python stuff is not available at all, so I suggest disabling CFFI for 42.2 i586 as well.

vyskocilm:
Hi  well I will test conditionals in prjconf then. Can you revert the last
commit please?

Dne 21. 6. 2017 11:32 dopoledne napsal uživatel "Luca Boccassi" <
notifications@github.com>:

PR #932, opened 2017-06-27
Problem: python package name has underscores only since f901cb9

vyskocilm:
Solution: fix paths in redhat packaging

@bluca - I did not find a similar section for Debian packaging. I'll let you to handle it ;-)


bluca:
The Debian build just calls pysetup from the directory so no change should be necessary

PR #937, opened 2017-07-10
Execute tests in the source directory. Some files needed by the tests…

diorcety:
… are relative to this directory

For example there is malamute with use "src/passwords.cfg"

### DIFF
--- a/zproject_cmake.gsl
+++ b/zproject_cmake.gsl
@@ -359,6 +359,7 @@ foreach(TEST_CLASS ${TEST_CLASSES})
     add_test(
         NAME ${TEST_CLASS}
         COMMAND $(project.prefix)_selftest --continue --verbose --test ${TEST_CLASS}
+        WORKING_DIRECTORY ${SOURCE_DIR}
     )
     set_tests_properties(
         ${TEST_CLASS}


bluca:
Please use the

```
Problem: x

Solution: do y and z to solve x
```

Format for commit messages

jimklimov:
This is why there are selftest-ro and selftest-rw directories. Many tests just can not be done in source directory and expect to have all files magically there, e.g. distchecks - files have to be EXTRA_DIST'ed into expected locations. This is what those directories do.

jimklimov:
Also, just in case: this fix makes the current dir for tests then"root" of the source? Not the "src" subdir in particular (that one would make tests behave differently with different tool suites)?

And regenerate zproject itself - travis is red ;(

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

bluca:
Yeah though I had seen that before - @diorcety could you please fix it?

diorcety:
Sûre. But i didn't understand what is the dix to do.

bluca:
Rather than setting the working directory to the srcdir, use selftest-ro or -rw like autools is doing, search for "selftest-r" in https://github.com/zeromq/zproject/blob/master/zproject_autotools.gsl

PR #941, opened 2017-07-12
Cmake and src/selftest-r{o,w} directories

diorcety:
#937


bluca:
@diorcety please use the following format for commit messages as requested by the C4 process:

https://rfc.zeromq.org/spec:42/C4/

```
Problem: x is broken

Solution: do y and z to fix x
```

diorcety:
? It is what i did

bluca:
Ah sorry, usually Github copies and pastes that as the PR title/description when there's only one commit so I got confused, my bad

Issue #952, opened 2017-08-04
New GSL needed to pass zproject regen job


jimklimov:
Per failed zproject selftest at https://travis-ci.org/zeromq/zproject/jobs/260882527 the job uses `GSL/4.1c Copyright (c) 1996-2016 iMatix Corporation` probably coming from packages by OBS, somehow provided via docker.

The latest GSL (and there were some fixes in the past days) identifies as `4.1.3`. I verified locally that self-regeneration with the new GSL does not change the sources (job passes).

Either the pre-built package (or docker image) should be bumped, or the job should build its own `gsl` parser.

bluca:
OBS has the newest build, not sure where that Docker comes from and how to update it - perhaps it should just be changed to use the OBS package like everything else?

jimklimov:
Maybe. I'm in mountains today so won't fix it tediously, but solution sounds trivial... I think .travis.yml already delivers gsl - so it's about the selfcheck job script. Maybe gitweb can manage ;)

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

bluca:
Fixed by https://github.com/zeromq/zproject/pull/953

PR #953, opened 2017-08-05
Problem: old gsl used to selftest zproject

jimklimov:
Solution: go from docker to packages

Addresses issue #952


jimklimov:
Can't open travis history at my current connection to see if older builds had more lines logged. Certainly there are more in my console, but maybe it's stderr vs stdout logging?

PR #961, opened 2017-08-15
Problem: .editorconfig is missing

jimklimov:
Solution: port the file from malamute to seed in other zprojects

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+# This is a skeleton created by zproject.
+# You can add hand-written code here.
+# See http://editorconfig.org/ and https://github.com/editorconfig/ for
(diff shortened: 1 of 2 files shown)


sappo:
Why add editor configuration?

jimklimov:
So people collaborating on the same project (and using compatible IDEs or editors) do not post conflicting indentation, etc. and so improve the coding style. At least, it seems to be the point of that effort from casual looking.

Granted, I did not use this yet, just saw it in malamute and thought it's both cool and productive for diverse-community projects :)

sappo:
Okay, as gsl files hate tabs this is a good thing :)

Am 15.08.2017 16:35 schrieb "Jim Klimov" <notifications@github.com>:

PR #963, opened 2017-08-29
Problem: memory leaks from indirect libgrypt dependency

vyskocilm:
Solution: ignore init calls of libgcrypt


jimklimov:
Just in case: shouldn't there be an explicit deinit at teardown instead?

I've hit similar issue in NUT recently,  where usage of libneon loaded and init'ed libxml2 which I had to un-init when unloading (ltdl'ed) libneon - otherwise it "leaked" some libxml2 state storage. This did not happen in persistently dynamically linked case - just for ltdl's, though.

PR #966, opened 2017-09-04
Problem: cffi binding is erroneous if method name equals keyword in python 2

sappo:
Solution: Check if the method name equals a known python 2 keyword and escape it with "_py" to prevent syntax errors.


bluca:
Thanks!

PR #974, opened 2017-09-28
Problem: gsl points to old imatix namespace

vyskocilm:
Solution: change url to zeromq organization


jimklimov:
@mvyskocil, @c-rack : merged before green; regen of zproject itself is needed

PR #977, opened 2017-10-02
Problem: Jenkinsfile does not "make distcheck"

jimklimov:
Solution: add the option similarly to Travis recipe; fix and mark a few nearby issues in this area

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,6 +37,13 @@ pipeline {
                 }
             }
         }
+        stage ('distcheck') {
+            steps {
+                timeout (time: 10, unit: MINUTES) {
+                    sh 'make distcheck'
+                }
+            }
+        }
     }
     post {
         success {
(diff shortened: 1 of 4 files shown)


jimklimov:
Well, Jenkins runs on a lot of distros with different packaging and other toolkits. So I guess there is no universal solution, although their concept of "tools" comes close (though may be geared/best-evolved towards Java projects). Maybe a solution could make some use of node labels, where Jenkins admins would list crucial packages or features needed for builds - e.g. presence of asciidoc, or lua, or libxml, or whatever in a particular worker image. Querying envvars or using shell-scripts is quite an option, but generally there is a Jenkins master that distributes jobs across a number of slaves - which each can have a different platform and OS, so node-labels are probably a better way to get onto expected environment in the first place.

In practice, nothing forbids starting with a sufficient development environment preinstalled and building listed known_project dependencies from scratch, like `ci_build.sh` does - including calling this same script as a build step (though without benefits of Jenkins2 logs split into views and marked with success-statuses for each separately defined step).

PR #981, opened 2017-10-10
Problem: need to branch the code according system endianess

vyskocilm:
Solution: call AC_C_BIGENDIAN for autotools to determine the endianess of the
system. This defines WORDS_BIGENDIAN macro on big endian systems.


bluca:
Could you please add a commit to regen, if you've got a minute?

vyskocilm:
Yes, I forget to do it all the time :-)

bluca:
Thanks!

PR #982, opened 2017-10-11
Problem: SELFTEST_DIR declarations pollute manual pages

vyskocilm:
Solution: make the documentation shorter, declare them before each selftest
method, convert to macros, so no dynamic string allocation or different
declaration for C and C++ are needed.

### DIFF
--- a/zproject_skeletons.gsl
+++ b/zproject_skeletons.gsl
@@ -296,36 +296,18 @@ $(actor.name:c)_actor (zsock_t *pipe, void *args)
 //  --------------------------------------------------------------------------
 //  Self test of this actor.
 
+// If your selftest reads SCMed fixture data, please keep it in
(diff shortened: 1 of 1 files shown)


jimklimov:
Macros can not be overridden at runtime like variables, e.g. via envvars (`getenv()`). But in practice we never came to require that possibility.

I believe the copypaste-friendly documentation like
````
-    // The following pattern is suggested for C selftest code:  
315    -    //    char *filename = NULL;  
316    -    //    filename = zsys_sprintf ("%s/%s", SELFTEST_DIR_RO, "mytemplate.file");  
317    -    //    assert (filename);  
318    -    //    ... use the filename for I/O ...  
319    -    //    zstr_free (&filename);  
320    -    // This way the same filename variable can be reused for many subtests.  
````

is still valid with macros as well, though?

PR #983, opened 2017-10-13
Problem: debian packages do not build in parallel

jimklimov:
Solution: it does not suffice to just `export DEB_BUILD_OPTIONS=parallel=8` or `MAKEFLAGS=-j8` or call `dpkg-buildpackage -j8` ... the `dh` command in the `debian.rules` file must have the `--parallel` flag as well - according to my experiment and to https://www.linuxquestions.org/questions/linux-software-2/parallel-jobs-with-dpkg-buildpackage-4175502264/

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -47,4 +47,5 @@ override_dh_auto_configure:
 
 %:
 	dh $@ \
+		--parallel \
 		--with autoreconf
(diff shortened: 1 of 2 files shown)


bluca:
Good catch - most of my stuff is compat 10 by now so I forgot:

```
If neither option is specified, debhelper currently defaults to --parallel in compat 10 (or later) and --no-parallel otherwise.
```

https://manpages.debian.org/testing/debhelper/debhelper.7.en.html

PR #985, opened 2017-10-18
Problem: Debian packaging not at recent compat level

jimklimov:
Solution: Use DH_COMPAT=10 to enable parallel build support by default


bluca:
@jimklimov sorry we need to revert this, it breaks builds on Debian 7 and Ubuntu 12.04 which have debhelper 9.20120909 and are still supported, we can switch in a year or so when 7 goes EOL and all Travis instances are moved on from 12.04

PR #988, opened 2017-11-07
Problem: can't install python3 cffi bindings on openSUSE

vyskocilm:
Solution: have a dependency on python, not python3


bluca:
Isn't the python 3 package called python3? It looks like from the logs: https://build.opensuse.org/build/network:messaging:zeromq:git-draft/openSUSE_Leap_42.3/x86_64/czmq/_log

vyskocilm:
Yes, there are python 2 and and python 3  packages. It's rusts any problem?

bluca:
So shouldn't the python 3 bindings depend on the python 3 package?

PR #990, opened 2017-11-08
Problem: there's no python3 on older rhels

vyskocilm:
Solution: distinguish python vs python3 build conditionals for rpmbuild

### DIFF
--- a/zproject_redhat.gsl
+++ b/zproject_redhat.gsl
@@ -51,6 +51,11 @@ register_target ("redhat", "packaging for RedHat")
 %bcond_with python_cffi
 %if %{with python_cffi}
 %define py2_ver %(python2 -c "import sys; print ('%d.%d' % (sys.version_info.major, sys.version_info.minor))")
+%endif
+
+# build with python3_cffi support enabled
+%bcond_with python3_cffi
+%if %{with python3_cffi}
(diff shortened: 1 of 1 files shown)


zproject_redhat.gsl
```diff
@@ -266,10 +284,15 @@ find %{buildroot} -name '*.la' | xargs rm -f
...
-python3 setup.py install --root=%{buildroot} --skip-build --prefix %{_prefix}
 %endif
 
+%if %{with python3_cffi}
+cd bindings/python_cffi
+export PKG_CONFIG_PATH=`pwd`
+python3 setup.py install --root=%{buildroot} --skip-build --prefix %{_prefix}
+%endif
 .endif
+
```
bluca: @vyskocilm just a nit, but any chance this empty line could be moved before the .endif above it? This way it won't add a new line when there's no python at all, like in zproject. See for example:

https://travis-ci.org/zeromq/zproject/jobs/299554278

vyskocilm:
Interesting, you can't remove line using phone. The backspace in editor
does nothing.

Dne 9. 11. 2017 11:52 dopoledne napsal uživatel "Luca Boccassi" <
notifications@github.com>:

PR #992, opened 2017-11-09
Remove scope around the use of utf8 converters 

bjornwennberg71:
Otherwise the internal utf8 storage will delete the buffer before it is used

### DIFF
--- a/zproject_nodejs.gsl
+++ b/zproject_nodejs.gsl
@@ -362,7 +362,9 @@ using namespace Nan;
     if (info [$(index () - 1)]->IsUndefined ())
         return Nan::ThrowTypeError ("method requires a `$(name)`");
 
-    $(c_type) $(c_name);
+    //$(c_type) $(c_name); // bjornw typedef - if using c_type, then you get 'int * major' but it needs to be 'int major'. later using the FromJust() returns an int
+    $(nan_type) $(c_name); 
+
 .           if type = "boolean"
 .               my.check = "IsBoolean"
 .               my.what = "Boolean"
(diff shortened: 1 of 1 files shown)


bjornwennberg71:
When you send a string using js that is longer than 354 bytes long (or thereabout) only garbage is sent.

Utf8String has 2 buffers. One is 1024 bytes long, and if the input-string exceeds this amount,
it will dynamically allocate a longer string. And it will deallocate this string in its dtor.
So, when the bindings uses the Ut8String it goes out of scope, crap happens.

BUGGY METHOD:
NAN_METHOD (Zstr::_send){
    Zsock *dest = Nan::ObjectWrap::Unwrap<Zsock>(info [0].As<Object>());
    char *string;
    if (info [1]->IsUndefined ())
        return Nan::ThrowTypeError ("method requires a `string`");
    else  if (!info [1]->IsString ())
        return Nan::ThrowTypeError ("`string` must be a string");
    else {
        Nan::Utf8String string_utf8 (info [1].As<String>());
        string = *string_utf8;
	// BUG HAPPENS HERE
	// Utf8String::dtor() happens and our 'string' points to crap
    }
    int result = zstr_send (dest->self, (const char *)string);
    info.GetReturnValue ().Set (Nan::New<Number>(result));

}

CORRECT METHOD:
NAN_METHOD (Zstr::_send){
    Zsock *dest = Nan::ObjectWrap::Unwrap<Zsock>(info [0].As<Object>());
    char *string;
    if (info [1]->IsUndefined ())
        return Nan::ThrowTypeError ("method requires a `string`");
    else  if (!info [1]->IsString ())
        return Nan::ThrowTypeError ("`string` must be a string");
    else {
        Nan::Utf8String string_utf8 (info [1].As<String>());
        string = *string_utf8;

        int result = zstr_send (dest->self, (const char *)string);
        info.GetReturnValue ().Set (Nan::New<Number>(result));
    }

PR #998, opened 2017-11-22
Problem: README.md not held up to date

jimklimov:
Solution: add a recipe to use gitdown for updating the file (still needs an explicit "make README.md" - not part of default build), and Travis CI to test for that

Known deficiencies:
* better make it part of `make all`
* better do this in common `configure.ac` and `Makefile.am` (and `zproject_autotools.gsl` so this can be spread to other projects)

After some practice with this approach, these improvements can be done at a later stage.


vyskocilm:
It looks like ci is broken, cannot install ImageMagick. Merging

jimklimov:
Well, broken means broken - could be a typo in new `.travis.yml` lines (seems it was), etc. Red ball means something ;)

Working on follow-up to get it green.

jimklimov:
Yay, the fixed code did catch the existing README mismatch errors via Travis : https://travis-ci.org/jimklimov/zproject/jobs/305878685

PR pending

PR #1000, opened 2017-11-22
Problem: Jenkinsfile initial support is too rudimentary

jimklimov:
Solution: implement the "TODO" features to support a real pipeline from Git to optional deployment (as implemented by a particular Jenkins deployment - not part of generally reusable FOSS project skeleton generated by zproject), and add an example skeleton for a deployment job that a particular Jenkins instance admin can set up and tailor to their reality.

Configuration-wise, this template should be compatible with earlier option names that early adopters of zproject Jenkinsfile support could have already used.


jimklimov:
CC @thalman

jimklimov:
Note: due to a known Jenkins issue, "the parameters aren't known until after the first run unfortunately" - so the first automatic run after discovery of a branch with Build Arguments requirement in a MultiBranch Pipeline fails. Afterwards one can do a Build with Parameters at least manually (or by committing something new), and then it works automagically.

Thinking of a few ideas to perhaps automate this... maybe in a follow-up PR.

PR #1008, opened 2017-11-25
Problem: Ruby bindings are very noisy around draft/deprecated functions

paddor:
Solution: DRY the FFI attachment of draft and deprecated functions using
two Ruby helper methods:

* ProjectName::FFI.attach_deprecated_function()
* ProjectName::FFI.attach_draft_function()


paddor:
Thanks!

PR #1010, opened 2017-11-26
Problem: Ruby bindings are fragile

paddor:
It assumes that all the stable functions are provided by an installed
library. But this assumption breaks as soon as the API contains a new
stable function that is not available in an already installed version of
that library. (See zeromq/czmq#1795)

Solution: gracefully handle any function that is not available in the
installed library, not only draft/deprecated ones. A placeholder
method is defined that will simply raise NotImplementedError.


paddor:
Thanks!

PR #1023, opened 2017-12-12
Allow static/shared builds for cmake generator similar to libzmq CMakeLists.txt

glaure:
Problem: It is not possible to build shared and static builds using the zproject_cmake template, (Or a static build ...)

Solution:
The changes create a CMakeLists.txt file very similar to the one libzmq uses.

Example: Features added to czmq
adds options CZMQ_BUILD_SHARED and CZMQ_BUILD_STATIC (both default on)
MSVC: create static and shared libs with the name: czmq-v120-mt-s-4_0_3.lib, czmq-v120-mt-4_0_3.dll
MSVC: export CZMQ_STATIC flag
Generate a valid CZMQ_VERSION variable
UNIX: shared: generate versioned so_name
UNIX: static: generate standard unversioned name
export include_dirs and compile flags to depending projects

I hope this does not break anything. This change was tested with czmq on windows and linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,12 @@ if (MSVC)
         PROPERTIES LANGUAGE CXX
     )
     set(MORE_LIBRARIES ws2_32 Rpcrt4 Iphlpapi)
+
+    if (MSVC_IDE)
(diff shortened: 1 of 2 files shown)


jimklimov:
Regenerate zproject itself please

PR #1042, opened 2018-02-01
Clang-format recipe support via CMake and Travis CI

jimklimov:
Actual work developed by @sigiesec - I just converted it into zproject generated code. His work allows to make sure the codebase conforms to the defined style, and if not - it generates a patch-file to fix the codebase (as part of its Travis job), so developers are not even required to have a local `clang` installation to benefit from this check and to make their commits conformant.

Note: with `clang-format` and `cmake` available on a local system, a developer (apparently) can `make clang-format-diff` in their codebase to fix it right away. I hope an equivalent solution for `autotools` fans appears soon, too; as well as a Jenkins integration.

UPDATE: Basic autotools integration has been added (discovery if the tool is present, and ability to use it from CLI `make`)


vyskocilm:
Hi, having code style check is neat +1! It can spot the violations

However I have a problem with this part

>  and if not - it generates a patch-file to fix the codebase (as part of its Travis job)

Please turn the automatic patching off by default. I don't think it's great idea to allow computers to touch the code without human intervention. Even @hintjens wrote the same in his Ten Rules for Good Code, even his perspective was more about the social and human aspect.

> Do not use automatic reformatters, as they hide mischief. A good style is worth applying by hand.
http://hintjens.com/blog:96

BTW: You've forgot to define the problem you're solving ;-)

bluca:
> Please turn the automatic patching off by default. 

Don't worry that's just in the output as guidance, it won't change the code itself :)

vyskocilm:
@bluca 👍  glad that I misread the description

jimklimov:
Technically, I believe executing that recipe, `make clang-format` in your workspace - preferably after committing your other recent work, would actually re-format the working checkout you have. Then you can revise it with `git difftool` (Meld being my X11 comparator of choice) and `add` and `commit` the result, or just `git checkout -f` to revert to the workspace as of the latest commit with intended changes.

PR #1045, opened 2018-02-02
Problem: #define format is not consistant

barraudl:
Signed-off-by: barraudl <lilianbarraud@eaton.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1383,4 +1383,4 @@ Before opening a pull request read our [contribution guidelines](https://github.
 
 ### This Document
 
-_This documentation was generated from Zproject/README.txt using [Gitdown](https://github.com/zeromq/gitdown)_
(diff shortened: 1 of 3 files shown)


sappo:
@barraudl `c_name` has already been formatted with `:c`. To tell gsl to print something as is just add `:` without a suffix.

barraudl:
@sappo Is the purpose of class.c_name is to be a class.name formatted with :c ? Or it's another way to call class.name ? (before class.name was allways formatted with :c so it is a bit redoundant in both case)

sappo:
@barraudl `class.c_name` is the classes name according to CLASS style guide which is achieved by applying the `:c` pretty printer. In other language bindings the class name cannot be formed with a simple pretty printer and for consistency they're named `c_name`, `java_name`, `python_name`, etc.

PR #1050, opened 2018-02-13
Problem: can't build rpm due not found man1 page

vyskocilm:
Solution: partially revert d27b30fea702341df3582f762277bda9b5befe78
Never ever mangle man1 documents, they must be the same as names of
mains itself, otherwise it will be totally confusing for end users.

### DIFF
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -2288,7 +2288,7 @@ generate_manpage ($name, $outp, $rootsrcdir);
 .   endfor
 .   man1 = ""
 .   for my.project.main where scope = "public"
-.      man1 += " $(name:c).1"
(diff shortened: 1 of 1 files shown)


zproject_autotools.gsl
```diff
@@ -2299,7 +2299,7 @@ $(project.GENERATED_WARNING_HEADER:)
 all-local: doc
 
 # Public programs ("main" tags in project.xml), auto-regenerated:
-MAN1 =$(man1:no)
+MAN1 =$(man1)
```
geraldguillaume: Do you really  want  "must be the same as names" ?
Because it is not the case.

BTW, you broke the compatibility with MAN3.
AFAIU, class name with uppercase produce lower case documentation name.

PR #1058, opened 2018-02-21
Allow projects without czmq dependency (part 1)

michal42:
The generated code expects that czmq pulls in a few headers and macro
definitions. This is a minimal fix for the generated selftest file in
C++ mode.

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -205,6 +205,9 @@ $(project.GENERATED_WARNING_HEADER:)
 
 #include "$(project.prefix)_classes.$(project.header_ext)"
 
+#include <stdio.h>
+#include <string.h>
+
 typedef struct {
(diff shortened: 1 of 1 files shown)


zproject_class.gsl
```diff
@@ -260,7 +263,7 @@ test_available (const char *testname)
 {
     test_item_t *item;
     for (item = all_tests; item->testname; item++) {
-        if (streq (testname, item->testname))
```
bluca: It would be good to check if this and other macros can be pulled in here too,in the private project header

michal42: I thought about it, but in the end I decided not to do it. If somebody does not want a particular dependency, let's not shove parts of the dependency down their throat regardless. We could define the macros in the *_selftest file directly, if you think it's worth it.

Issue #1054, opened 2018-02-19
Usage of Visual Studio configuration names not consistent


hambre:
In file zproject_vs20xx.gsl the visual studio project files are genarated with the following build configuration names: DebugDLL, ReleaseDLL, DebugLTCG, ReleaseLTCG, DebugLIB, ReleaseLIB.
The generated build file build.bat use the following build configuration names: DynDebug, DynRelease, LtcgDebug, LtcgRelease, StaticDebug, StaticRelease (line 705-749).
So building fails with msbuild complaining that the configuration attempted to build does not exists.

bluca:
Fixed by https://github.com/zeromq/zproject/pull/1060

PR #1060, opened 2018-02-21
fix configuration names in build.bat

hambre:
Building on windows by calling build.bat fails as it tries to build
projetc configurations which do not exists in the visual studio solution
or project files. This commit changes the configuration names in the genereated
build.bat file to be in sync with the visual studio solution and project
file.


hambre:
This refers to issue #1054

PR #1071, opened 2018-03-09
Problem: debian copyright is always from 2015

jimklimov:
Solution: make the project.starting_year a configurable value

### DIFF
--- a/packaging/debian/copyright
+++ b/packaging/debian/copyright
@@ -2,7 +2,7 @@ Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: zproject
 
 Files: *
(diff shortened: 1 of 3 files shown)


bluca:
Could you please regen the readme?

As a future enhancement, perhaps the license.xml file could be improved with more fields, to be able to generate structured data. So the copyright year range could be an additional element in the license file itself, for example. Other things could be short license text, etc.

jimklimov:
Ok, another PR upcoming - I'll put the regen'ed README there.

jimklimov:
In https://github.com/zeromq/zproject/pull/1072/commits/514a03da06babfb0be0fc1398f16529f538c9f87 of #1072 I moved the starting_year into license.xml, seems to have just worked :)

PR #1072, opened 2018-03-09
Refactor the clang-format-check-CI to use in both CI recipes

jimklimov:
Also fix the starting_year from #1071 comments


jimklimov:
The capital Z in readme seems to have been an artifact of current directory name :\

bluca:
Yes, there's plenty of that with zproject unfortunately - for example generated zyre bindings includes depend on the czmq directory name in the same parent

PR #1078, opened 2018-04-05
Problem: can not customize configure.ac final words

jimklimov:
Solution: extend support of optional acinclude.m4 with ability to define an AX_PROJECT_LOCAL_HOOK_FINAL routine with things to do just before AC_OUTPUT

Verified in a project of ours having one or both of these macros defined - the m4 `ifdef` performed as expected, so no breakage.

Pro tip: When debugging, beware to not `echo` that we execute a macro by name from inside that macro - or you get an infinite loop ;)


bluca:
missing regen

jimklimov:
Ack, had it locally but not committed :)

PR #1084, opened 2018-04-30
Problem: rpm dependencies are not upgraded by yum/zypper

vyskocilm:
Solution: make the dependencies tight, so enforce %{version}-%{release}
must be the same. This is what will users want to do anyway


bluca:
Looks like it's failing:

```
gsl/4 M: Building packaging for RedHat (redhat)
(zproject_redhat.gsl 114) Undefined expression: my.abi_ver
```

Also please regenerate zproject

vyskocilm:
will revork it, closing now

PR #1088, opened 2018-05-23
Problem: zproject does not generate cmake package configs

gabm:
Solution: add this functionality

by default, all dependencies are treated as `PUBLIC` because in zproject they are not handled separately at the moment. Another limitation is that the users maybe want to modify the Config.cmake.in file. Here we should allow the inclusion of a local file - just like `CMakeLists-local.txt`...

tested with czmq and zyre


gabm:
This code is partially taken from the libzmq CMakeLists.txt ..

PR #1092, opened 2018-05-31
Problem: ZMQ-style compact streq() replaced by strcmp()==0 monstrosity

jimklimov:
Solution: Restore the pre-Feb2018 coding pattern (ease regeneration of less dynamic projects) and retain the benefits of that patch (do not require builds with czmq headers).
    
CCing @bluca @michal42 whose work is reverted here (note especially Luca's patch - was the original cause a syntax curiosity of MacOS compilers, or a logical error in the string comparison)?

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -215,9 +215,6 @@ $(project.GENERATED_WARNING_HEADER:)
 
 #include "$(project.prefix)_classes.$(project.header_ext)"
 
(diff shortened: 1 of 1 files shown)


bluca:
Sorry but I don't really remember...
In general, all of the CZMQ header should move to zproject, it's just a historic accident that it's there, but it doesn't belong in a public header of a library

PR #1102, opened 2018-07-23
Revert "Problem : a project.xml without czmq dependency is not buildable"

karolhrdina:
1) **Provided solution does not work**

Even if we pretend for a moment, that declared "problem" is a problem, then this pull request does not provide a working solution:
https://gist.github.com/karolhrdina/b3b702d529d66502698dd509ea1906e8

This gist has two files, `project.xml` -  a minimal zproject xml without czmq dependency, `output` - output of doing gsl project.xml; ./autogen.sh && ./configure && make, which fails as seen.

2) **Provided 'Problem' statement is not a problem**

> zproject is a project that makes it cheaper to build other projects that use ZeroMQ.

^^^ is a quote from Hintjens himself in a article http://hintjens.com/blog:79 presenting zproject to the public for the first time. 
`zproject` is not supposed to work without `czmq`; This fact is obvious in every part of gsl template for CLASS or ACTOR elements; `czmq` is integral to `zproject`, since czmq is the high level C binding for libzmq.

### DIFF
--- a/zproject_skeletons.gsl
+++ b/zproject_skeletons.gsl
@@ -177,7 +177,7 @@ struct _$(actor.name:c)_t {
 static $(actor.name:c)_t *
 $(actor.name:c)_new (zsock_t *pipe, void *args)
 {
-    $(actor.name:c)_t *self = ($(actor.name:c)_t *) malloc (sizeof ($(actor.name:c)_t));
+    $(actor.name:c)_t *self = ($(actor.name:c)_t *) zmalloc (sizeof ($(actor.name:c)_t));
     assert (self);
 
     self->pipe = pipe;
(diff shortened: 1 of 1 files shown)


bluca:
I think the fact that zproject depends on czmq is mostly an "accident" of history: it existed before it, and had a bunch of utilities that have little to nothing to do with zmq itself but are of general usefulness in C projects, so it was included like this.

Therefore it would be fine to make zproject work without czmq: I think there's already folks using it that way. In fact, I've long thought that most of the stuff, macros and checks and functions that are in https://github.com/zeromq/czmq/blob/master/include/czmq_prelude.h really don't belong there, and should be moved to zproject as private code. ENOTIME to take care of that though...

karolhrdina:
brutal ENOTIME here as well, however off top of my head, how would you
suggest to support e.g. <actor ...> element without czmq?

vyskocilm:
You don't, zproject can be changed to add an implicit dependency on czmq if
actor is used. I agree with Luca, czmq prelude should be moved to zproject.
There are quite some nice helpers...

Dne út 24. 7. 2018 11:41 uživatel karolhrdina <notifications@github.com>
napsal:

karolhrdina:
I would prefer if we wouldn't mislead users by adding a dependency that is
not listed in the project template (which would happen if we add it after
re-generation and the user does not change the project.xml).

My suggestion would be to do it the other way around:

IF you don't list czmq as a dependency => fail in gsl template on 'actor'
element with a message. Just like when we had the error with include
directory not created.

What do you guys think?
Other than that, i agree and now i can see the benefit of moving prelude to
zproject.

Thanks for the idea.

K.

PR #1105, opened 2018-07-31
[DO-NOT-MERGE] Problem: implicit dependency on czmq

vyskocilm:
Solution: move some of czmq_prelude.h definitions to _library.h. I was
focused on making sure generated project can build, so most of things
weren't copyied. What is in ZPROJECT_PRELUDE now

1. All ANSI and platform dependent includes, those comes with very
   little cost for compiler nowadays and allows us to use `printf`/`puts`
   safely in generated code
2. `bool`/`true`/`false` definition on Windows and `stdbool.h` include
   elsewhere, because `bool` is required by standard _test method
   declaration
3. `streq`/`strneq` macros, those are part of generated code
4. `zmalloc` routine, as is called from generated code

In order to maintain backward compatibility, ZPROJECT_PRELUDE is skipped
if czmq_prelude.h exists. This creates new problem that those
definitions lives in czmq_prelude.h and are generated by zproject as
well. However this is out of scope of this commit.

cc: @bluca @karolhrdina @geraldguillaume 
references: https://github.com/zeromq/zproject/pull/1101
references: https://github.com/zeromq/zproject/pull/1102

Note this is quite big change, however solves the problem welI placed all definitions to one place because 'bool' declaration was needed for public includes and it is much easier to have all prelude stuff in one place. The downside is that evolution of generated preludes is impossible, because the lowest component will win. And of course this creates additional problems of duplicating some parts of code between czmq and zproject.

### DIFF
--- a/zproject_python.gsl
+++ b/zproject_python.gsl
@@ -157,6 +157,9 @@ endfunction
 
 function method_annotation (method)
     fn = "lib.$(class.name:c)_$(method.name:c)"
+    if method.private ?= 1
+        fn="#lib.$(class.name:c)_$(method.name:c)"
+    endif
     for my.method.return
         restype = return.python_type
     endfor


bluca:
This can't go in a public header, otherwise we are back at square one. It should be in a private header, included only in internal source code, in the src/ directory. That way each project has its own and they don't stomp on each other.

vyskocilm:
@bluca I think I can move 3-4 to `src/_classes.h`, however 1. is harmless and 2. is there because bool is required by generated _test functions in public header files. Sound fine to you?

bluca:
Other projects might do weird and quirky things that break with the simple inclusions of headers. Think glibc VS BSD libc and so on. Good libraries must not impose these things on downstream users. It should all be private.

PR #1107, opened 2018-08-02
Problem: implicit dependency on czmq

vyskocilm:
Solution:
if project is not czmq or does not depends on it
1. move `__WINDOWS__` definition to `_library.h`
2. include basic ANSI headers mandatory for generated code in private `_classes.h`
3. include common idioms streq/strneq/zmalloc in `_classes.h`
4. fail if `actor` is specified, but `czmq` is not in the dependency list

cc: @bluca @karolhrdina @geraldguillaume
references: #1101
references: #1102
references: #1105 

This is leaner and smaller version of previous attempt. Tested manually with minimal project.xml(s) and with czmq.

### DIFF
--- a/zproject_class.gsl
+++ b/zproject_class.gsl
@@ -83,7 +83,7 @@ $(project.GENERATED_WARNING_HEADER:)
     $(PROJECT.PREFIX)_MAKE_VERSION($(PROJECT.PREFIX)_VERSION_MAJOR, $(PROJECT.PREFIX)_VERSION_MINOR, $(PROJECT.PREFIX)_VERSION_PATCH)
 
 // czmq_prelude.h bits
-.if project.prefix <> "czmq"
+.if project.prefix <> "czmq" & project.has_czmq <> 1
 #if !defined (__WINDOWS__)
 #   if (defined WIN32 || defined _WIN32 || defined WINDOWS || defined _WINDOWS)
 #       undef __WINDOWS__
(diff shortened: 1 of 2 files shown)


zproject_class.gsl
```diff
@@ -508,6 +519,45 @@ typedef struct _$(class.c_name:)_t $(class.c_name:)_t;
 .endfor
 
 //  Internal API
+.if project.prefix <> "czmq" & project.has_czmq <> 1
+// common definitions and idioms from czmq_prelude.h, which are used in generated code
+#if ! defined(__CZMQ_PRELUDE_H_INCLUDED__)
```
bluca: I think there's a potential race condition here - if a source file includes this header files first, and then czmq_prelude.h, there will be duplicate definitions (some compilers might barf).
So, given this should be an alternative to it, I think __CZMQ_PRELUDE_H_INCLUDED__ should be defined to avoid it

bluca:
Thanks, this is a good first step - it's too bad that the test API uses a bool forcing the stdbool include, but given it's already there I don't think we can get away with changing it

vyskocilm:
Bool has been introduced in c99, so expecting nearly 20 year old features
for new C projects is not that bad thing, imho.

For c++ we have enforced at least c++11 anyway.

Dne 2. 8. 2018 12:44 odp. napsal uživatel "Luca Boccassi" <
notifications@github.com>:

Thanks, this is a good first step - it's too bad that the test API uses a
bool forcing the stdbool include, but given it's already there I don't
think we can get away with changing it

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/zeromq/zproject/pull/1107#issuecomment-409885149>, or mute
the thread

bluca:
Yeah that's a fair point - although Windows does its usual Windows thing...

zproject_class.gsl
vyskocilm: Hi @bluca - I missed this comment.

I don't think there is race as external dependencies are always included before Internal API section. And if someone will include czmq.h AFTER ... well, then he should simply put czmq into project's dependencies. In this case following code won't be generated.

PR #1109, opened 2018-08-08
Problem: header files listed twice confuse "install"

jimklimov:
Solution: verify that classes and actors are not named same as the project, their headers should differ too
Note: gsl abort() currently seems flawed in not returning a non-zero exit code, but that is a separate issue

Note: the initial issue was that we saw duplicate header filenames listed in `include/Makefile.am` for common list and draft additions. Digging into the issue produced this fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ zproject's `project.xml` contains an extensive description of the available conf
     description := A short description for your project (optional)
     email := The email address where to reach you (optional)
     repository := git repository holding project (optional)
-    unique_class_name := "0"|"1" (optional, defaults to 1) As a failsafe, forbid naming agents or classes same as the project itself (can cause conflicts in generated header filenames). Disable (set to 0) only in legacy projects that can not regenerate otherwise, and try to fix those.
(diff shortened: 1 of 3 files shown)


bluca:
Won't this break zyre?

bluca:
https://github.com/zeromq/zyre/blob/master/project.xml#L16

jimklimov:
Probably will, and per cursory analysis - it is already a bug there: by my understanding, both items generate header files, a projectname.h and a classname.h. Not sure if into same directories, so whether it is a confusing inconvenience or direct conflict, but incorrect it does feel.

Also, if zyre class were marked draft, it would generate a include/Makefile.am that lists a zyre.h twice and gets our original issue.

bluca:
Is there any way to fix it without breaking compatibility?

jimklimov:
I am not sure really, because apparently this is retaining compatibility with something broken...
That said, there's always an option of adding an option :) so a particular project can disable this check...

jimklimov:
Hopefully the new toggle and https://github.com/zeromq/zyre/pull/600 address this concern

bluca:
Given it breaks existing things, it should default to disabled

bluca:
There might be private projects using it, so printing a warning for now is the right thing, and we can change it in the future. Thanks

PR #1113, opened 2018-08-14
Problem: dependencies checked out into workspace root

jimklimov:
Solution: avoid potential conflicts and simpify gitignore maintenance by dedicating a tmp-deps subdirectory for the checkout and build of dependency components.

Note: I tested this with a few projects that have Travis for autotools, and ported the fix but not tested for CMake and for bindings. One example is at https://github.com/42ity/fty-example/pull/24/files

### DIFF
--- a/builds/check_readme/ci_build.sh
+++ b/builds/check_readme/ci_build.sh
@@ -27,10 +27,13 @@ cd "${REPO_DIR}" || exit
 PATH="`pwd`:$PATH"
 export PATH
 
+rm -rf tmp-deps
+mkdir tmp-deps
+
 echo "=== Install gitdown"
-git clone https://github.com/zeromq/gitdown.git gitdown
 mkdir -p ~/bin
-( cd gitdown && ./install-wrapper ~/bin ) || exit $?
+( cd tmp-deps && git clone https://github.com/zeromq/gitdown.git gitdown ) \
+&& ( cd tmp-deps/gitdown && ./install-wrapper ~/bin ) || exit $?
 PATH="$PATH:~/bin"
 export PATH
 which gitdown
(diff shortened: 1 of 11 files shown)


bluca:
Does this work for example when regenerating zyre which needs czmq in the same parent dir?

jimklimov:
Good question, should be easy to check... but i'm on the road this week and on vacation in the next two. I'll try to test but not certain asap.

Why is zyre such an oddball? ;)

FWIW, usecase with these checkouts (or tarballs) in pure generated code does not guarantee presence of third party sources at all - if suitable packages are installed in the build env, sources are not needed nor built. So if zyre requires czmq sources, it maybe uses some customized routine?

bluca:
It's not specific to zyre - due to the way some zproject bindings work, you need the dependencies checked out in the same parent directory for the re-generation to work correctly

jimklimov:
Well, same *parent* directory is not touched by this changeset. It is about checking out code into current workspace with root of the project itself, mostly in CI usecase.

I tried to regenerate zyre checked out alone into a /tnp and saw that some zproject files for zproto, czmq, ...  were included from /usr/local/share/zproject/... and there were no complaints about no other prohect sources near or under this dir. Also seems zyre's zproject was not regenerated in over a year ;)

bluca:
Ok great, thanks for checking (zyre was regenerated a few days ago though?)

jimklimov:
Indeed, there are some hardships in https://github.com/zeromq/zyre/pull/601 - not all due to tmp-deps/ however. I've addressed some, but believe that the check_zproject part depends on upstream or packaged zproject, so there's not much I can do for testing it otherwise.

Maybe I can make a WIP branch that would reference my fork for Travis needs... but it sounds too complicated for a rest from vacationing :\

jimklimov:
Hopefully the check_zproject script itself was an overkill easy to fix... testing started but laptop battery goes off :\

bluca:
thanks for checking

jimklimov:
By the way, zproject does produce a warning about missing dependencies in expected location. Perhaps it should abort() there?

PR #1116, opened 2018-08-26
Clean up some more fallout from #1113

jimklimov:
The `check_zproject` script verifies that an expected set-up of a developer workstation includes dependency checkouts in proper locations (into same parent dir as the checked project itself). Thus, it should not use `tmp-deps/` after all.

Also, generated projects' code should not refer to `tmp-deps/` either - if the dependencies' files are to be used directly without a checkout, it may be an unfortunately intimate type of dependency, but one that should be reproducible in a casual developer's environment manually per README instructions, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,11 +71,13 @@ install_manifest.txt
 Testing/
 
 # Repositories downloaded by CI integration scripts
-gsl/
-gitdown/
+tmp-deps/
+###gsl/
+###gitdown/
 
-# Manually tracked: the czmq source is pulled as a test target
-czmq/
+### Manually tracked: the czmq source is pulled as a test target
+###libzmq/
+###czmq/
 
 # Travis build area
 tmp/
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -29,8 +29,10 @@ esac
 
 case "$BUILD_TYPE" in
 "default"|"distcheck")
-    mkdir tmp
+    rm -rf tmp tmp-deps
+    mkdir -p tmp tmp-deps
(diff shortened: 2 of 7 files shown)


bluca:
One of the tests fails:
```
gsl/4 M: E: please checkout libzmq into ../libzmq
Command exited with non-zero status 1
```

jimklimov:
Logs at https://travis-ci.org/zeromq/zproject/jobs/420788093 lead to an interesting chicken-and-egg problem: the `autogen.sh` which is executed early in the build of a project (test of czmq regenerated with the tested zproject code to be built built from scratch) essentially requires that libzmq is available so a GYP file can be made for it.

If the directory is never checked out (per manually maintained script in this case, or per packaging availability in common use-case), the GYP file can not be properly regenerated automatically. If the prereqs are missing, lines would be dropped from the regenerated GYP file making it into not what the original developer intended. While I can and will amend this for CI of zproject itself (explicitly check out libzmq before czmq), I am not sure there is a good general solution for mixing checking-out and not-checking-out of dependency repos.

Probably they should always be checked out (perhaps via `autogen.sh` of each generated project? but adding an internet dependency in each run?), maybe into `tmp-deps/` if not available in a parent dir of a workspace (e.g. `tmp-deps` of a higher-in-the-stack project, possibly traversing layers of the parent dirs?), or the dependency has to be avoided via re-engineering somehow.

I do have an interest in reasonably solving this dependency of some zproject-driven projects on others for another quest, so constructive discussion of possible ways and their pro's and con's is welcome ;)

PR #1120, opened 2018-10-07
[WIP] problem: zproject missing a way to configure/enable/install language bindings

wesyoung:
solution: modify zproject_autotools.gsl to provide basic configure flags for python bindings and drop an "make bindings" option in `src/Makemodule.am` when a project is created.

this is a prototype using python, if there's a more elegant way of doing this i'm open to feedback.. i made sure not to check in updates to `configure.ac` and `src/Makemodule.am` even though this gsl updates those and they are already checked into the project (i can re-do that if we want them updated, i didn't know what kind of chaos that might create).

it's specific to python but i wanted to lay the groundwork for other language bindings so users could:

```
$ configure --enable-bindings-python=yes && make && sudo make install && make bindings
```

and the correct language bindings would end up on the underlying OS. this would help reduce the amount of overhead higher level bindings (such as pyzyre, etc) would need to maintain.

feedback welcome..


bluca:
looks fine to me, want to send the regenerate as well?

wesyoung:
ack- good comments. my only concern (and the reason i didn't try to generalize it just yet), in practice most people might not (probably won't?) have all the non-obvious language deps they need for the bindings they're installing. so maybe if anything, the "make bindings" would only install things you've configured. i could go either way, but chose to be more specific in the beginning and see where it leads.. (easier for others to follow, no accidental breakage, etc).

does that make sense?

wesyoung:
it's all good- i've been wrapped up in other stuff myself. thank you for the feedback- this is why i wanted to let it simmer a bit.

wesyoung:
i *think* i did the regen correctly(?), i'll try to incorporate some of the feedback in future PRs.. we can bat this around see if/how it evolves..

bluca:
can you rebase instead of merging please?

PR #1126, opened 2018-11-13
Problem: travis recipe allows to only do(1) or skip(0) distcheck

jimklimov:
Solution: add an option to generally skip it, but run it in a special test case allowed to fail (2) as a project approaches maturity


jimklimov:
Example travis run here: https://travis-ci.org/jimklimov/fty-example/builds/454720171

PR #1135, opened 2018-12-12
[JNI] Load C library from within the jar

sappo:


sappo:
@hgourvest this might be useful for you.

hgourvest:
@sappo it's a good thing, thanks!

PR #1161, opened 2019-02-21
Problem: timer units installed into /usr/lib

jimklimov:
Solution: fix path to match other units (/lib)
TODO: Honor configured @systemdsystemunitdir@ somehow?


bluca:
the install file is debian-specific and they (and derivatives) use /lib so it should be fine - it's the makefile install that should use pkg-config to get the unitdir path

PR #1168, opened 2019-04-21
Problem: Findlibzmq.cmake doesn't work on Win32

myd7349:
Findlibzmq.cmake doesn't work out-of-box on Win32.
libzmq dll/lib built with MSVC is named using the Boost convention(something like libzmq-mt-4_3_1.lib, libzmq-mt-4_3_1.dll): https://github.com/zeromq/libzmq/blob/master/CMakeLists.txt#L1147-L1173
So find_library (LIBZMQ_LIBRARIES NAMES zmq HINTS ${PC_LIBZMQ_LIBRARY_HINTS}) will not work.

This patch tries to solve the `libzmq not found` error on Win32 with CMake:
https://github.com/zeromq/czmq/issues/1972#issuecomment-484842328
https://github.com/zeromq/czmq/issues/2004

### DIFF
--- a/zproject_cmake.gsl
+++ b/zproject_cmake.gsl
@@ -661,11 +661,79 @@ find_path (
     HINTS ${PC_$(USE.PROJECT)_INCLUDE_HINTS}
 )
 
+.       if use.libname = "libzmq"
+if (MSVC)
+    # libzmq dll/lib built with MSVC is named using the Boost convention.
+    # https://github.com/zeromq/czmq/issues/577
+    # https://github.com/zeromq/czmq/issues/1972
+    if (MSVC_IDE)
+        set(MSVC_TOOLSET "-${CMAKE_VS_PLATFORM_TOOLSET}")
+    else ()
+        set(MSVC_TOOLSET "")
+    endif ()
+
+    # Retrieve ZeroMQ version number from zmq.h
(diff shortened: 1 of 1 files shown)


myd7349:
I'm not sure if it the best way to solve this problem. If there is a better one, just let me know.

bluca:
Does it have to be specific to libzmq? Shouldn't it be a general solution for all libraries?

myd7349:
@bluca Thanks!
AFAIK, only a little libraries(ZeroMQ, Boost) built on Windows use this naming strategy. So there is no need to be a general solution currently.

zproject_cmake.gsl
```diff
@@ -661,11 +661,79 @@ find_path (
...
+        REGEX "#define ZMQ_VERSION_(MAJOR|MINOR|PATCH)")
+    foreach(ver ${zmq_version_defines})
+        if(ver MATCHES "#define ZMQ_VERSION_(MAJOR|MINOR|PATCH) +([^ ]+)$")
+            set(ZMQ_VERSION_${CMAKE_MATCH_1} "${CMAKE_MATCH_2}" CACHE INTERNAL "")
+        endif()
+    endforeach()
+
+    set(_zmq_version ${ZMQ_VERSION_MAJOR}_${ZMQ_VERSION_MINOR}_${ZMQ_VERSION_PATCH})
+
+    set(_zmq_debug_names)
```
myd7349: I made a mistake here: Forget to delete these two lines.

PR #1169, opened 2019-05-09
Problem: python_cffi binding can't handle double pointers

sphaero:
Solution: if the api says the argument is a reference create a ** pointer to it


sphaero:
There's still a problem with destroying messages but this needs other addressing

PR #1191, opened 2019-06-21
Problem: test for C++11 (so g++-4.9) is unavoidable now

jimklimov:
Solution: make the check for std::regex (assuming it means the sufficient
level of C++11 in GCC) optional with --enable-gcc-std-regex=no and by
default enabled if gcc-4.9 is requested by the project's "classfilename" tag.

A build with this option set to 'no' should compile in some of the earlier
toolchains, such as the default gcc-4.8 in "trusty" VMs on Travis, but the
result would not necessarily function (gcc-4.8 and earlier could include the
API routines for supporting the standard, but not really implement it).

Primary intended use-case:

If Travis CI tests of some projects "use" a project that by itself
requires C++11, but do not really use this aspect of it in their own
unit-tests, this is sufficient for "trusty" machines with older GCC.
In this case, the "use" tag for such project can be amended with
````
    <add_config_opts>--enable-gcc-std-regex=no</add_config_opts>
````
tags to avoid configure-time failures in Travis CI builds of the
whole dependency chain (unpackaged mode).

### DIFF
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -1278,7 +1278,7 @@ int main() {
 #endif
   return result ? EXIT_SUCCESS : EXIT_FAILURE;
 }
-], AC_MSG_RESULT([ok]), AC_MSG_ERROR([test for std::regex failed]) )
+     ], AC_MSG_RESULT([ok]), AC_MSG_ERROR([test for std::regex failed]) )
      AC_LANG_POP([C++])
     ]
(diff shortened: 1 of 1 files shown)


jimklimov:
Note: not tackled now, but maybe it would make sense for someone's use-case to reshuffle the code to set this or that default value for this new option (checking for std:regex) based on presence of gcc as the compiler, on requesting gcc-4.9 in project.xml etc. and make available for all compilers the test based on this setting (not hidden for just the gcc-4.9+ requirement as it is now).

If going this way, also note to *not* emit the warning about skipping the test if the codebase (project.xml) did not declare that the feature is actually required from the compiler.

PR #1196, opened 2019-08-29
Problem: can't drop python from my system because of zproject

vyskocilm:
Solution: do not install `mkapi.py` and `fakecpp`. To be honest I would
personally frop them from zproject as I am not sure it even works
nowadays :)


bluca:
If the problem is python2, what about converting them to python3?

vyskocilm:
@bluca I believe it is forward compatible with python3 - it use print_function for instance. However I personally did not use it in years and have no idea it the tool works and if it's useful to anyone else. I'd use the removal as a check if someone will complain or not :)

PR #1204, opened 2019-10-10
Problem: default starting_year of 2015 in projects

jimklimov:
Solution: if a project is being just generated, and there
is already an `<include filename="license.xml">` in `project.xml`
but no `license.xml` (verbatim) file yet, create one with the content
of `<starting_year>` (set to today's year) and a `<license>` tag
with a reminder to define an actual license and regenerate.

Do not touch existing project license files though.

Signed-off-by: Jim Klimov <EvgenyKlimov@eaton.com>

Example output:
````
fty-example$ cat license.xml
<starting_year>2019</starting_year>
<license>
LICENSE FOR THIS PROJECT IS NOT DEFINED!

Copyright (C) 2019- by fty-example Developers &lt;zmqbob@company.com&gt;

Please edit license.xml and populate the 'license' tag with proper
copyright and legalese contents, and regenerate the zproject.

LICENSE FOR THIS PROJECT IS NOT DEFINED!
</license>
````

PR #1205, opened 2019-10-10
Problem: too easy to re-release a project with obsolete Copyrights

jimklimov:
Solution: Take legalese seriously. During regeneration, check
if the current year is mentioned (from, until, listed, whatever)
in a line with the "Copyright" word or some of the ways it can
be written in the XML `<license>` tag.

If the year is not listed in any such line (or no such lines are
there), warn the user in at least the build-logged messages, and
also pause for 5 sec so the interactive people can break, fix and
regenerate.

Loosely related to #1204 but not as invasive :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,8 +223,11 @@ zproject's `project.xml` contains an extensive description of the available conf
     name := The name of your project (optional)
     description := A short description for your project (optional)
     email := The email address where to reach you (optional)
+    url := The website or similar resource about the project or its ecosystem (optional)
     repository := git repository holding project (optional)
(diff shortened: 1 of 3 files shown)


bluca:
Having a warning is fine, but there's no need for a pause - after the bern convention the "copyright YYYY-YYYY" line is actually not legally required almost anywhere in the world, and is purely informative. Copyright is implicit and automatic, and with git there is enough metadata to satisfty any lawyer/court.

jimklimov:
Good point, thanks... I'll see if I can make it a lever then to not impact general population but have it in our project to be cleaner and give the sources a sense of having been maintained recently :)

bluca:
Yeah having warnings is fine, it's good maintenance to keep those dates updated.

jimklimov:
Here, this should cover all the bases and satisfy all tastes :)

jimklimov:
Missed one commit...

jimklimov:
Ironically and notably, zproject itself would have the warnings as of current state (no years in its own `license.xml`)

Issue #1200, opened 2019-09-24
Cmake Xcode project fails


sphaero:
Trying to open a zproject library in OSX results in errors ons builds:
```
PhaseScriptExecution CMake\ PostBuild\ Rules libsphactor.build/Debug/sphactor.build/Script-478FADA0B354431F8E16B705.sh
    cd /Users/ect/src/libsphactor
    /bin/sh -c /Users/ect/src/libsphactor/libsphactor.build/Debug/sphactor.build/Script-478FADA0B354431F8E16B705.sh

echo "Creating symlinks"
Creating symlinks
/usr/local/Cellar/cmake/3.14.0/bin/cmake -E cmake_symlink_library /Users/ect/src/libsphactor/Debug/libsphactor.0.0.1.dylib /Users/ect/src/libsphactor/Debug/libsphactor.0.dylib /Users/ect/src/libsphactor/Debug/libsphactor.dylib
CMake Error: failed to create symbolic link '/Users/ect/src/libsphactor/Debug/libsphactor.0.dylib': no such file or directory
CMake Error: cmake_symlink_library: System Error: No such file or directory
CMake Error: failed to create symbolic link '/Users/ect/src/libsphactor/Debug/libsphactor.dylib': no such file or directory
CMake Error: cmake_symlink_library: System Error: No such file or directory
make: *** [sphactor_buildpart_0] Error 1
```

The makefile generated by cmake expects files at certain places which aren't there:
```
# Generated by CMake, DO NOT EDIT
# Custom rules for sphactor
.SUFFIXES: 
all: \
	sphactor_buildpart_0


sphactor_buildpart_0: 
	echo "Creating symlinks"
	/usr/local/Cellar/cmake/3.14.0/bin/cmake -E cmake_symlink_library /Users/ect/src/libsphactor/Debug/libsphactor.0.0.1.dylib /Users/ect/src/libsphactor/Debug/libsphactor.0.dylib /Users/ect/src/libsphactor/Debug/libsphactor.dylib
```
The  /Users/ect/src/libsphactor/Debug/ dir doesn't exist. 

Anybody had success with Xcode?

aaronvark:
I tried to get it to build, and came up with the following:

If I manually create the Debug folder, the symlink creation does not fail. But then the build still fails because it tried to find a file with the name "0.0.1.dylib", instead of the ".dylib" or "0.dylib" files that it has symlinked.

```
Ld osxbuild/Debug/sphactor_selftest normal x86_64
[...]

PR #1208, opened 2019-10-16
Fix linker issue in cmake -G Xcode projects

aaronvark:
re #1200

Tested on OSX Mojave (10.14.6) and XCode 11.1.

Previously a generated Xcode project would not compile because an incorrectly named .dylib file was being generated (it defaulted to "0.dylib", but our project.xml defined "0.0.1.dylib"). After some searching this might actually be a non-compliance with Apple guidelines (they want only major versions in .dylib, and minor/patch in config or something.

This edit uses sources more directly, and bypasses the whole issue.


bluca:
this breaks re-using the same object files for both builds

Issue #1217, opened 2020-02-01
Making releases based on specific dependency branches is complicated


jimklimov:
Zproject and its library of `known_projects.xml` really helps get a multi-component project including its git/rpm/deb/... dependencies off the ground... as long as you are following "master" (or equivalent default) git branches of the projects you are using, and in case of known projects - take them from the common upstream repository.

Due to peculiarities of our own project's release and maintenance strategy, this approach did not work our of the box - but luckily, we can override `project.xml` as it is being parsed by adding `<use` tag contents to inject the needed repository (for forked upstreams) and "release" branch name. We can even nest the use-in-use to ensure(*) that prerequisites get built on e.g. travis in correct order, when building from scratch and without packages.

* Nesting has caveats, sometimes top-level "use" tag order matters. This is a bug likely, but separate from this issue.

With 50+ (and growing) components comprising our "distributed monolith" product codebase, it becomes quite tedious to branch and update all the "release" attributes and regenerate zproject's to take into account that they need to checkout that bolted-down source version whenever needed. The overall set of components and their interdependencies also evolve, so preparing a release was also our last chance to make sure every component's `project.xml` trees made sense on the given month. Painfully, it took a couple of man-weeks to branch in order (and sanity-check) the tree of components involved in a release.

We theorized about some solutions, and it boiled down to a few possibilities, most with some downsides:
* fork zproject to have our own known_projects.xml pointing to our own forks - suffers from using also a correct branch of zproject to also point to correct "releases". And maintenance hell for keeping yet another fork up to date.
* extend upstream zproject for using site-custom known_projects.xml if available (and/or load one like we load `licence.xml`), as an override to the upstream one... this does make sense, may fit many use-cases, but requires that developers doing the release also have the correct variant of the override XML installed.
* Generate the (includable?) cumbersome tree of dependencies for a project as part of its `gsl zproject` run, by walking recursively the projects it refers to (if they have a project.xml) - lots of git traffic during project generation and/or needed smartness about already available local workspaces to avoid a new requirement to be online to regenerate a zproject.
[...]

PR #1218, opened 2020-02-01
Recursive CI builds for dependencies

jimklimov:
This PR implements the last option listed in #1217 - allowing to minimize the `project.xml` considerably, keeping just the top-level `use` tags for dependencies whose APIs are directly used by this component, e.g. https://github.com/42ity/fty-asset/pull/120/files#diff-c0a33b20a5bf94bd7bd5c012bf96f702

As detailed in #1217, this approach is not a magic silver bullet and gotta have use-cases that do not fit its idea and/or implementation.

It is recognized that this solution originated for our pain point, being "Travis Linux builds our unpackaged ecosystem from scratch with gcc/automake" (so there is no similar handling for MacOS or CMake, or non-travis builds). If the idea picks up, followers are welcome to fill the gaps.

It also introduces the notion of a new optional magical file `.ci_global_release` that would cause the CI source-builds of this component to request either certain branch or its currently checked out branch for the dependencies.

Cheers and kudos go to my colleagues. I did the initial analysis outlined in the #1217 issue, and was skeptical about the particular route they took, but it seems to have panned out well and simplified our release process in the end. Now we want to share back (hopefully removing the need to have a fork of zproject), and to get additional sanity review from the community :-)

I am not sure how critical it was so did not include in the PR, but there was also removal of the hardcoded reference to non-default branch of `libsodium` (`stable` rather than `master`) from `known_projects.xml` so our components could refer to a particular branch themselves:
* https://github.com/42ity/zproject/commit/92c1d54c377412233dd414d8b15bea5e91c84c38
* https://github.com/42ity/zproject/commit/66f5e0600e0e1743c398f66e774625fc204db957

IIRC it was a problem because of ZMQ ecosystem pulling it in, and `czmq` hardcoded as being required by `zproject` generated codebases at some point.

### DIFF
--- a/zproject_travis.gsl
+++ b/zproject_travis.gsl
@@ -719,7 +719,10 @@ if \
     if [ -e ci_dependencies.sh ]; then
         PROPAGATED_BRANCH="`git branch | grep \* | cut -d ' ' -f2`"
         DEFAULT_BRANCH="`git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'`"
-        if [ "x$PROPAGATED_BRANCH" = "x$PROPAGATED_BRANCH" ]; then
+        if [ "x$REQUESTED_BRANCH" = "x" ]; then
+            echo "`date`: INFO: Building prerequisites of '$(use.project)' using ci_dependencies.sh the default branch..." >&2
+            ($CI_TIME source ./ci_dependencies.sh)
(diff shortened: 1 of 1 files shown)


jimklimov:
CC @perrettecl @boricj @stark-dev

jimklimov:
Note: looking at the linked commit reducing the project.xml, I saw it also removed the `test` keywords; I am not sure OTOH if it was required by the current build sequence change or happened by poor luck. Certainly, referencing dependencies that have known capabilities (exposed  needed symbols) makes sense where possible.

PR #1221, opened 2020-02-05
Resync from opensource

jimklimov:
This includes the (other history path of) our upstreamed recursive-CI build support, as well as other patches, to stay on par with opensource/master.


jimklimov:
wrong repo sorry

PR #1222, opened 2020-02-10
Dockerfile fixes and improvements 

sappo:


zproject_known_projects.xml
```diff
@@ -33,7 +33,7 @@
         </use>
     -->
 
-    <use project = "libzmq" prefix = "zmq" debian_name = "libzmq3-dev" redhat_name = "zeromq-devel"
+    <use project = "libzmq" prefix = "zmq" debian_name = "libzmq5-dev" redhat_name = "zeromq-devel"
```
bluca: the old name is the correct one

bluca: https://sources.debian.org/src/zeromq3/4.3.2-1/debian/changelog/#L157

sappo: I see libzmq5-dev is used in latest Ubuntu though. Maybe we need to differentiate between them.

bluca: it's not used, it's just a metapackage - https://tracker.debian.org/media/packages/z/zeromq3/control-4.3.2-1

sappo:
@bluca I removed the commit changing the debian package.

PR #1241, opened 2020-03-10
Enforce assert in release mode AND travis ci build fix for windows

sappo:


zproject_class.gsl
```diff
@@ -505,6 +505,16 @@ $(project.GENERATED_WARNING_HEADER:)
 //  External API
 #include "../include/$(project.header:)"
 
+//  Asserts check the invariants of methods. If they're not
```
bluca: This block should probably be above the previous include - otherwise if NDEBUG is set, and another library or one of its dependencies includes assert.h, this block will be a no-op due to the include guards

sappo:
Okay, I'll move it.

Luca Boccassi <notifications@github.com> schrieb am Di., 10. März 2020,
21:41:

zproject_class.gsl
sappo: Done

Issue #1265, opened 2020-09-30
PROBLEM: Platform checks for AI_V4MAPPED not present in cmake


mjvankampen:
I made a fix in czmq to handle AI_V4MAPPED not being present on netdb.h for every system (just added QNX, ref. https://github.com/zeromq/czmq/pull/2119). 

Problem: These defines (HAVE_ANDROID etc.) are not in the CMakeFile generated by zproject. 

Solution: 
Various solutions exist, I'm willing to make a PR but need some guidance on the best way to go.
1. Add HAVE_PLATFORM defines to cmake
1. Add HAVE_NOT_AI_V4MAPPED to cmake and czmq
1. Add HAVE_AI_V4MAPPED to cmake and autotools (and others?) and czmq

What would be the preferred way to implement this? I think adding option defines to cmake minimizes impact, but is the least clean. Option 3 is the nicest has the most impact, and option two is a middle way.

I would prefer option 2 I think as I am not that familiar with autotools and I can just use https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html. Of course option 1 is possible too but not really the cmake way to go.

sappo:
@mjvankampen thanks for bringing up this issue.

I went for another option. Zproject provided for both autotools and cmake local hooks to customize their script with project related content.

I used the check_symbol_exists function in cmake and their counterpart in autotool to make them set an define. See zeromq/czmq#2121.

We could enhance zproject though to declare those symbol check in the project.xml and have them generate the appropriate cmake and autotool code. What do you think? Most dev probably know at most one option and that way it's more obvious that this symbol check exists.

mjvankampen:
I think that is the best way to go! Most akin to option 3. I think this makes the code in the projects that use zproject more clear and robust.

sappo:
Fixed by #1267

PR #1267, opened 2020-10-06
Problem: Symbol check requires knowledge about multiple build systems

sappo:
Solution: Declare the check in the project.xml and let zproject generate
the details.

### DIFF
--- a/zproject_autotools.gsl
+++ b/zproject_autotools.gsl
@@ -1042,6 +1042,13 @@ AC_TYPE_UINT32_T
 AC_C_VOLATILE
 AC_C_BIGENDIAN
 
+.for project.check_symbol_exists
+.if first()
+# Check for symbols (function, variable, or macro)
+.endif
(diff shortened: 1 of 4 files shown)


sappo:
Solves #1265

zproject_autotools.gsl
```diff
@@ -1042,6 +1042,13 @@ AC_TYPE_UINT32_T
 AC_C_VOLATILE
 AC_C_BIGENDIAN
 
+.for project.check_symbol_exists
+.if first()
+# Check for symbols (function, variable, or macro)
+.endif
+AC_CHECK_DECLS([$(symbol:)], [], [], [[#include <$(header:)>]])
+.endfor
+
```
bluca, line 1051: moving the newline inside the for, behind .if last() ensures no empty new line added everywhere

sappo: Yep #1268

PR #1270, opened 2020-11-27
Problem: Issue with generated android build script.

sailaja-nagireddy:
Solution: Update the android gsl to add the location from android ndk while looking for readelf binary.


sappo:
Thanks

PR #1307, opened 2022-10-05
Problem: Android build scripts silently fail.

stephan57160:
This change only requires the 2 other proposed fixes.


stephan57160:
**Note 1:**
Compilation is OK but we're still unable to use the final .jar.

**Note 2:**
From the begining, I have to apply 2 additional patches:
- a temporary fix for --secure-transport and CURVE.
- a temporary fix to skip GRADLE automatic tests. They loop, and I'l still unable to understand why.
- the final .jar reports an error with missing ZMQ, despite present in the .jar file.

PR #1311, opened 2022-10-16
Problem: Android APP fails to load ZMQ since NDK r25.x

stephan57160:
Same cause than explained in
https://github.com/zeromq/libzmq/pull/4437

Solution: Fix invalid LDFLAGS.

Notes:

- Still fails on physical devices.


stephan57160:
Same (or very similar) fix than already merged to LIBZMQ.

Info from my co-worker:

- Possibly fails since NDK 24.

- Working on some physical devices (armv7), then, my feeling is that we face a problem witth LLVM on AARCH64.
I think I read something about that, some months ago, but unable to find it again ...

PR #1335, opened 2024-04-05
Problem: Fix misspelling & out-of-sync zproject

BHare1985:
Solution: Fix the misspelling in the .gsl and re-run it.

This was initially found in czmq where a committed changed a generated file and the misspelling came back once I generated. This should fix it upstream so that it fixes all downstream. 

CZMQ should not need re-generated after this merge because it already had the misspelling fixed. See https://github.com/zeromq/czmq/pull/2283


BHare1985:
I noticed @tblattner  made a change in the zproject-generated file that this is now reversing: https://github.com/zeromq/zproject/pull/1330

Do you have context on what issue you were having @tblattner? as generally I see warnings that the variable names should be the same as package. Perhaps the solution to whatever issue you were having is related to changing `Findgsl.cmake` instead ?

BHare1985:
Radio silence from @tblattner.  I would like to go ahead and push this through @bluca or know what you would like to see to get it in.

sappo:
Why did gsl_FOUND change to lower case?

BHare1985:
> Why did gsl_FOUND change to lower case?

It is a result of running `gsl project.xml`

Changing it to all caps was a manual modification from https://github.com/zeromq/zproject/pull/1330 which this effectively reverts.

